PR #31, opened 2023-01-23
feat: Support custom serializers

manzt:
This is a WIP in progress at providing at declarative API for supporting custom serializers in **anywidget**. Custom `serializers` are a fairly advanced widget concept, but can be useful for handling custom data-types (specifically binary data). 

This PR implements support via another (optional) ESM export, `serializers`:

```javascript
export const serializers = {
  my_value: {
    deserialize: (value, manager) => /* ... */,
    serialize: (value, manager) => /* ... */,
  }
}

  let deserialized = view.model.get("my_value");
  /* ... */
}
```

which extends the `DOMWidgetModel.serializers` like when defining traditional Jupyter Widgets. 

## More trouble than it's worth?

The main challenge with supporting custom `serializers` is that the Jupyter Widgets framework relies on serializers being defined statically on a unique model that extends from `DOMWidgetModel`.  All **anywidget**'s derive from the _same_ `AnyModel` class, so we can't just modify `AnyModel` itself (because one widget could override the serializers for another). Further, we don't have access to the `_esm` until after the model has been initialized, so we can't know the serializers until after initialization.

The key to the proposed solution, is that following Widget Model initialization, the Jupyter Widgets framework accesses the statically defined serializers via the model _instance_ constructor (i.e., `model.constructor`) for all [comm messages](https://github.com/jupyter-widgets/ipywidgets/blob/74774dae5becb9f4781c3438a5fece1f8ca60415/packages/base/src/widget.ts#L265). Through the magic of JavaScript [Proxies](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy), we can inspect and intercept access to `model.constructor`, effectively enabling the re-definition of properties on `AnyModel` for a specific `model` _instance_ in a controlled manner.

```javascript
let overrides = { serializers };
this.model.constructor = readonly_proxy(AnyModel, overrides);
// this.model now users the above serializers in `_deserialize_state`, rather than whatever was defined on `AnyModel` originally

// merged serializers 
this.model.constructor = readonly_proxy(AnyModel, { serializers: { ...AnyModel.serializers, ...serializers } });

// original behavior
this.model.constructor = AnyModel; // restore behavior
```

While this solution works currently, there are several issues that make me nervous:

- We can't register serializers until _after_ the initial state has been deserialized. This means that we need manually deserialize _some_ of the model state the first time we encounter a model instance.

- It's a hack. This implementation relies heavily on an implementation detail in Jupyter Widgets, which could change and then what are we to do.

- Manually performing serialization/deserialization isn't that bad and would be more stable.

## Alternatives?

Originally I thought that custom `serializers` was a very low-level mechanism of transforming data to/from Python (i.e., handling bytes), but upon further inspection  it's just an additional hook provided by Jupyter Widgets to _further_ transform an already JSON-decoded object. Essentially, serializers are just syntactic sugar for:

```javascript
const deserialize = (value, manager) => /* ... */;
const serialize = (value, manager) => /* ... */;

// export const serializers = { my_value: { deserialize, serialize } };

let deserialized = await deserialize(view.model.get("my_value"), view.model.widget_manager));

let serialized = serialize(value, view.model.widget_manager);
view.model.set("my_value",  serialized);
```

Which is fairly easy to write helpers for in user code. It begs the question of whether we should put in all this effort to have a similar declarative API or just suggest users write little serializer utilities for accessing state. Odds are they have a helper like:

```javascript
function getCount() {
	return view.model.get("count");
}
```

and is it too much to ask for it to be changed to

```javascript
function getCount() {
	let serialized = view.model.get("count");
	return deserializeCount(serialized, view.model.widget_manager);
}
```

If needed?

Finally, a limitation of Jupyter Widgets serializers is that `deserialize` can be asynchronous but `serialize` cannot. This means if you require async transformations for `serialization` you have to manually serialize/deserialize the state like the snippet above (i.e., you can't use the builtin serialization mechanim).

### DIFF
--- a/src/widget.js
+++ b/src/widget.js
@@ -1,9 +1,13 @@
 // @ts-check
 import { name, version } from "../package.json";
 
+/** @typedef {import("@jupyter-widgets/base").DOMWidgetView} DOMWidgetView */
+/** @typedef {import("@jupyter-widgets/base").ISerializers} ISerializers */
+
 /**
  *  @typedef AnyWidgetModule
- *  @prop render {(view: import("@jupyter-widgets/base").DOMWidgetView) => Promise<void>}
+ *  @prop render {(view: DOMWidgetView) => Promise<void>}
+ *  @prop [serializers] {ISerializers}
  */
 
 /**
@@ -92,6 +96,33 @@ async function load_esm(esm) {
 	return widget;
 }
 
+let anywidgetSymbol = Symbol("anywidget");
+
+/**
+ * Jupyter Widgets define custom `serializers` on the Model class statically.
+ * All **anywidget** users derive from the same `AnyModel`, so we can't allow
+ * users to mutate `AnyModel` (otherwise there could be conflicting state names)
+ *
+ * Creating a proxy for the `model.constructor` allows us to inject serializers
+ * on top of those statically defined for the Model.
+ *
+ * @param {typeof import("@jupyter-widgets/base").DOMWidgetModel} Model
+ * @param {ISerializers=} serializers
+ */
+function createModelProxy(Model, serializers) {
+	/** @type {ISerializers} */
+	return new Proxy(Model, {
+		get(target, prop, receiver) {
+			// tag our proxy
+			if (prop === anywidgetSymbol) return true;
+			// intercept serializers
+			if (prop === "serializers") return serializers;
+			// allow original object to handle the rest
+			return Reflect.get(target, prop, receiver);
+		},
+	});
+}
+
 /** @param {typeof import("@jupyter-widgets/base")} base */
 export default function (base) {
 	class AnyModel extends base.DOMWidgetModel {
@@ -102,6 +133,11 @@ export default function (base) {
 		static view_name = "AnyView";
 		static view_module = name;
 		static view_module_version = version;
+
+		constructor(attributes, modelOptions) {
+			super(attributes, modelOptions);
+			this._initial_state = attributes;
+		}
 	}
 
 	class AnyView extends base.DOMWidgetView {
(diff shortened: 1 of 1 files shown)


manzt:
Example:

```python
import anywidget
import traitlets
import ipywidgets as ipw

class DatePicker(anywidget.AnyWidget):
    _esm = """
    export const serializers = {
        value: {
            deserialize(value) {
                if (value == null) return null;
                let date = new Date();
                date.setUTCFullYear(value.year, value.month, value.date);
                date.setUTCHours(0, 0, 0, 0);
                return date;
            },
            serialize(value) {
                if (value == null) return null;
                let date = {
                    year: value.getUTCFullYear(),
                    month: value.getUTCMonth(),
                    date: value.getUTCDate(),
                };
                return date;
            }
        }
    }
    export async function render(view) {
        let picker = document.createElement("input");
        picker.setAttribute("type", "date");
        picker.addEventListener("change", e => {
            view.model.set("value", picker.valueAsDate);
            view.model.save_changes();
        });
        function update_value() {
            let value = view.model.get("value");
            picker.valueAsDate = value;
        }
        update_value();
        view.listenTo(view.model, 'change:value', update_value);
        view.el.appendChild(picker);
    }
    """
    value = ipw.trait_types.Date().tag(sync=True, **ipw.trait_types.date_serialization)

import datetime
picker = DatePicker(value=datetime.date(year=2023, month=2, day=3))
picker
```

image

nvictus:
> an additional hook provided by Jupyter Widgets to further transform an already JSON-decoded object

Yeah, I agree that it doesn't seem worth it to engineer this in with a potentially fragile monkey patch.

I can see the motivation if you want to send raw bytes from Python to the client, but if it always receives JSON you might as well complete your "deserialization" in user code.

manzt:
> always receives JSON

The caveat is that the Jupyter Widgets framework "magically" sends binary data to the client. When a dictionary is encoded to JSON with `bytes` or `memoryview`, the Jupyter Widgets framework sends the binary data separately, and then the deserialized object in JavaScript is a binary [`DataView`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView) for that key.

For example,

```python
import anywidget
import traitlets
import numpy as np

def to_json(*args, **kwargs):
    ar = np.array([1, 2, 3], dtype="f4")
    return {
        "text": b"Hello, world.",
        "arr": { 'view': memoryview(ar), 'dtype': str(ar.dtype),  'shape': ar.shape }
    }

class BinaryWidget(anywidget.AnyWidget):
    _esm = """
    export async function render(view) {
        let { text, arr } = view.model.get("value");
        console.log(new TextDecoder().decode(text.buffer)); // Hello, world.
        console.log(new Float32Array(arr.view.buffer)); // Float32Array[1, 2, 3]
    }
    """
    value = traitlets.Any().tag(sync=True, to_json=to_json)

BinaryWidget()
```

manzt:
Closing for now. We could revisit in the future.

PR #32, opened 2023-01-24
chore: add license to pyproject.toml

manzt:


nvictus:
FYI, `license = { file = ... }` has a weird outcome on PyPI, where it copied the entire file to the PyPI metadata and made my license badge icon fail, so I switched to `license = { text = "BSD-3-Clause" }`

image

PR #42, opened 2023-01-31
docs: enable DocSearch

manzt:
**anywidget** was accepted for [DocSearch program](https://docsearch.algolia.com/docs/docsearch-program/)! This PR enables their search on the website. It's appears that it's fine to keep the Algolia API keys in the repo (Astro and Vite both do this).


manzt:
https://user-images.githubusercontent.com/24403730/215651188-4e9674b5-dcf3-41e5-8f23-9ede15f395f0.mov

PR #45, opened 2023-02-01
Raise error when instantiating invalid AnyWidget subclass

tlambert03:
in exploring the mechanics of this library, I was breaking all the rules :joy: and created an AnyWidget subclass without `_esm` or `_modules`.  In which case you see the following javascript error:

```
Failed to create view for 'AnyView' from module 'anywidget' with model 'AnyModel' from module 'anywidget'
TypeError: Cannot read properties of undefined (reading 'startsWith')
    at is_href (http://localhost:8888/nbextensions/anywidget/index.js?v=20230201171813:7:14)
    at load_esm (http://localhost:8888/nbextensions/anywidget/index.js?v=20230201171813:48:7)
    at AnyView.render (http://localhost:8888/nbextensions/anywidget/index.js?v=20230201171813:76:26)
    at async http://localhost:8888/nbextensions/jupyter-js-widgets/extension.js?v=20230201171813:2:742339
```

Obviously: one should just "rtfd" here, but this PR provides a more explicit runtime error.

```pytb
ValueError: Cannot instantiate AnyWidget without defining one of the following class attributes: ('_esm', '_module')
```

Since `_esm` is effectively an abstract class attribute, another option here (as opposed to raising in the `__init__`), would be to use `__init_subclass__`:

```py
class AnyWidget(ipywidgets.DOMWidget):
    ...
    def __init_subclass__(cls) -> None:
        if not any(hasattr(cls, attr) for attr in _ESM_CLASS_ATTRS):
            raise ValueError(
                "AnyWidget subclass must defining one of the following "
                f"class attributes: {_ESM_CLASS_ATTRS}"
            )
```

the main difference there is that the _immediate_ subclass of `AnyWidget` must have `_esm`... which could possibly restrict users making their own base classes. So this PR only errors at instantiation


manzt:
Thanks for the PR :) I guess another option would be to provide a default for `_esm` that doesn't do anything. 

Screen Shot 2023-02-01 at 6 55 31 PM

tlambert03:
Yeah I think a default no op would  be good too.  Can close this if you want!

tlambert03:
ok, changed it to this:

Screenshot 2023-02-02 at 11 25 37 AM

(where the link is to the docs getting started page)

thoughts?

anywidget/widget.py
```diff
@@ -3,6 +3,21 @@
...
+        let url = "https://anywidget.dev/en/getting-started/";
+        let p = Object.assign(document.createElement("p"), {
+            innerHTML: '<strong>Dev note</strong>: ' +
+            `<a href='${url}' target='blank'>Implement an <code>_esm</code> attribute</a>` +
+            ` on AnyWidget subclass <code>${view.model.get('_anywidget_id')}</code>` +
+            ' to customize this widget.'
+        });
+        root.appendChild(p);
+        view.el.appendChild(root);
+    }
```
manzt, line 19: Slight nit, we don't really need to create these elements with. I think this renders correctly if you just assign the `view.el.innerHTML` directly (the back-ticks string syntax supports multi-lines):

```suggestion
DEFAULT_ESM = """
export function render(view) {
  console.log("Dev note: No _esm defined for this widget:", view);
  let url = "https://anywidget.dev/en/getting-started/";
  view.el.innerHTML = `<p>
    <strong>Dev note</strong>:
    <a href='${url}' target='blank'>Implement an <code>_esm</code> attribute</a>
    on AnyWidget subclass <code>${view.model.get('_anywidget_id')}</code>
    to customize this widget.
  </p>`;
}
"""

Comment (manzt) [reply anywidget/widget.py:19]: image

File: anywidget/widget.py [hunk 2]
```diff
@@ -22,6 +37,9 @@ def __init__(self, *args, **kwargs) -> None:
             for k in ("_esm", "_module", "_css")
             if hasattr(self, k) and not self.has_trait(k)
         }
+        # show default _esm if not defined
+        if all(i not in anywidget_traits for i in ("_esm", "_module")):
+            anywidget_traits["_esm"] = t.Unicode(DEFAULT_ESM).tag(sync=True)
```
manzt, line 42: I think this would override `_esm` if the trait is explicitly registered via the `HasTraitsMetaclass`, e.g.,

```python
class CustomWidget(anywidget.AnyWidget):
    _esm = traitlets.Unicode(" ... ").tag(sync=True) # `anywidget_traits` will not include `_esm` or `_module` since `self.hastrait("_esm")` is true
```

Maybe we could move this logic after `self.add_traits(**anywidget_traits)`:

```python
if all(not self.has_trait(i) for i in ("_esm", "_module")):
    self.add_trait(_esm=t.Unicode(DEFAULT_ESM).tag(sync=True))
```

manzt: I think the `test_basic` unit test should be failing...

manzt: Ahh actually the unit tests don't assert the ESM is correct. Would you mind adding:

```diff
def test_basic():

    ESM = """
    export function render(view) {
        view.el.innerText = "Hello, world";
    }
    """

    class Widget(AnyWidget):
        _esm = t.Unicode(ESM).tag(sync=True)

    w = Widget()

    assert w.has_trait("_esm"
++  assert w._esm == ESM

def test_legacy():

    ESM = """
    export function render(view) {
        view.el.innerText = "Hello, world";
    }
    """

    class Widget(AnyWidget):
        _module = t.Unicode(ESM).tag(sync=True)

    w = Widget()

    assert w.has_trait("_module")
++  assert w._module == ESM


++ def test_missing():
++     class Widget(AnyWidget):
++       ...
++
++    w = Widget()
++    assert w.has_trait("_esm")
++    assert w._esm == DEFAULT_ESM
```

manzt (approved):
Looks great! Just some minor comments. I think we should add a unit test.

anywidget/widget.py
```diff
@@ -4,19 +4,17 @@
...
+  console.log("Dev note: No _esm defined for this widget:", view);
+  let url = "https://anywidget.dev/en/getting-started/";
+  view.el.innerHTML = `<p>
+    <strong>Dev note</strong>:
+    <a href='${url}' target='blank'>Implement an <code>_esm</code> attribute</a>
+    on AnyWidget subclass <code>${view.model.get('_anywidget_id')}</code>
+    to customize this widget.
+  </p>`;
+}
+"""
```
manzt, line 17:
```suggestion
```

PR #46, opened 2023-02-02
[draft] example with psygnal instead of ipywidgets

tlambert03:
hey @manzt 

this is a super basic (and still very broken) example of removing the python-side ipywidgets & traitlets requirements (using psygnal instead).  will require `pip install psygnal pydantic`. 

what's working:

- rendering of the initial widget using anywidget's js view
- _updating_ of the js WidgetModel whenever the python side model changes

what's broken:

- for some reason, you need to re-trigger a repr in order for the python model update to get shown in the view (so the js model isn't triggering a js view re-render for some reason yet)
- the view doesn't update the python model.  I'm still digging through traitlets and the `Comm` object to try to figure out how that all got connected.  I tried listening to `on_msg` ... but it didn't seem to do anything when the counter is clicked).  I can't tell if the JS state mutation is turning into a ZMQ message or not.

----

(it will be easy enough to remove the pydantic requirement as well later)


manzt:
@tlambert03 I'm so sorry about this... but the reason it's not working is because the JS code snippet doesn't register any update handlers. It doesn't faithfully setup callbacks to update whenever the model changes 😟

manzt:
Still some bugs but initially it seems to work with a frontend that correctly renders and subscribes to updates

https://user-images.githubusercontent.com/24403730/216470016-4d55c1ad-a003-464b-a54e-bfecbb08752d.mov

tlambert03:
> I'm so sorry about this... but the reason it's not working is because the JS snippet you used from examples/ doesn't register any handlers for when the model changes

I tried a few... and I feel like I tried the same snippet as an existing Anywidget subclass (successfully) and a Displayable subclass (unsuccessfully)... but maybe not?  Anyway that would be great!  Will take a look this weekend.  Thanks!

tlambert03:
> Still some bugs but initially it seems to work with a frontend that correctly renders and subscribes to updates
> 
> 
> 
> https://user-images.githubusercontent.com/24403730/216470016-4d55c1ad-a003-464b-a54e-bfecbb08752d.mov
> 
> 
> 
> 

Nice!

manzt:
Pushed a PR to your branch :) 

> I tried a few... and I feel like I tried the same snippet as an existing Anywidget subclass (successfully) and a Displayable subclass (unsuccessfully)... but maybe not? Anyway that would be great! Will take a look this weekend. Thanks!

Yeah, the notebooks don't do any assignment from the Python side though for (Counter.ipynb and Counter-htm.ipynb). Without the `on_msg` handler the python never got updated from the front-end either.

With my PR, they all work just like AnyWidget subclasses!

tlambert03:
🎉🎉🎉

tlambert03:
I continuing in breaking out the bits of functionality here, I made a working proof of principle on the decorator idea (#13)

this should now work:

```python
esm = """
    view.model.save_changes();
    });
    view.model.on("change:value", () => {
    });
    view.model.on("change:value", (...args) => {
    
        console.log(...args)
    })
    view.el.appendChild(btn);
}    
"""

@anywidget(esm=esm)
class Counter:
    value: int = 0
```

tlambert03:
@manzt ... I added one more pattern in the `_descriptor.py` module.  I _think_ it's my favorite so far,  API is similar to the decorator example above but everything is as lazy as possible, and it modifies the original object as little as possible (adding a single `_repr_mimebundle_` attribute).  i tried to document the module reasonably well to explain what's going on, but happy to step through it in person too

anywidget/_util.py
```diff
@@ -0,0 +1,123 @@
...
+    return Comm(
+        target_name="jupyter.widget",
+        metadata={"version": version},
+        buffers=[],
+        data={
+            # temporarily hard-code the model name and module
+            "state": {
+                "_model_module": "anywidget",
+                "_model_name": "AnyModel",
+                "_model_module_version": "0.1.0",
```
manzt, line 116: TODO: Right now the anywidget frontend code and the python package are published together with the same version. So `_model_module_version` and `_view_module_version` should be `anywidget.__version__`

PR #49, opened 2023-02-05
feat: add MimeBundleDescriptor pattern, for more library agnostic connection between python model and JS view

tlambert03:
closes #46

After discussion with @manzt, this PR represents the "essence" of what anywidget is doing, and allows it to work without depending on ipywidgets (though ipykernel is still required), and opens the door to a nicer end-user API using, for example, dataclasses or pydantic or attrs for defining a model, and psygnal (or traitlets, or any observer pattern) for synchronizing the front-end model/view and backend-model.

`MimeBundleDescriptor` is an object that implements the [descriptor pattern](https://docs.python.org/3/howto/descriptor.html), that will lazily open a `Comm` channel with the front-end only once the `_repr_mimebundle_` attribute is accessed (or whatever name it is assigned to on the class).  It attempts to detect a serialization pattern on the object, as well as an observer pattern.

At its core, (but least "attractive" looking), the pattern looks like this:

```python

ESM = """
    view.model.save_changes();
  });
  view.model.on("change:value", () => {
  });
  view.el.appendChild(btn);
} 
"""

class SomeClass:
    # added by this PR... this object manages the Comm
    # by default, it looks for observer patterns, but here we override that
    _repr_mimebundle_ = MimeBundleDescriptor(_esm=ESM, autodetect_observer=False)

    # bare minimum state getter
    # we can autodetect dataclasses and pydantic, etc...
    # but this method provides an general escape hatch
    def _get_anywidget_state(self):
        return {'value': self.value}

    # bare minimum observer pattern
    # here we  add an attribute, and call `send_state` when it changes
    _val = 0

    @property
    def value(self):
        return self._val

    @value.setter
    def value(self, val):
        self._val = val
        self._repr_mimebundle_.send_state({'value'})
```

at its "prettiest", using dataclasses for state getting/setting, and [psygnal](https://psygnal.readthedocs.io/en/latest/dataclasses/) for the observer pattern:

```python
@evented
@dataclass
class T1:
    _repr_mimebundle_ = MimeBundleDescriptor(_esm=ESM)
    value: int = 0
```

*edit*: or with traitlets:

```python
import traitlets

class Foo(traitlets.HasTraits):
    _repr_mimebundle_ = MimeBundleDescriptor(_esm=ESM)
```

to implement a decorator like #13 ... a very simple decorator would just look like this:

```python
def anywidget(esm: str):
    def _decorator(cls):
        setattr(cls, '_repr_mimebundle_', MimeBundleDescriptor(_esm=esm))
        return cls
    return _decorator
``` 

leaving you with a user API of:

```python
@anywidget(esm=ESM)
@evented
@dataclass
class T1:
    value: int = 0
```

~no tests yet... coming~ ... test coverage complete now


manzt:
Sweet 🤩 ! I was basically able to recreate the ipwidgets mechanism using traitlets:

```python
import traitlets

class Counter(traitlets.HasTraits):
    _repr_mimebundle_ = MimeBundleDescriptor(_esm=ESM, autodetect_observer=False)

    def notify_change(self, change): # traitlets
        if self.trait_metadata(change.name, "sync"):
            self._repr_mimebundle_.send_state({change.name})
        super().notify_change(change)

    def _get_anywidget_state(self):
        return self.trait_values(sync=True)
```

tlambert03:
excellent!  let's add that here (around [here](https://github.com/manzt/anywidget/pull/49/files#diff-72dfcbefe84cf7ce063c02cbf56f45f69fa2adb2ec3b49624d1f39d7381f5665R504), just like we look for psygnal)

tlambert03:
ok, I added traitlets support as well, this now works

```python
import traitlets

class Foo(traitlets.HasTraits):
    _repr_mimebundle_ = MimeBundleDescriptor()
```

tlambert03:
this is all set for review now

anywidget/_descriptor.py
```diff
@@ -0,0 +1,615 @@
...
+
+if TYPE_CHECKING:  # pragma: no cover
+    import traitlets
+    import pydantic
+    import psygnal
+
+    from ipykernel.comm import Comm
+    from typing_extensions import TypeGuard
+
+    from ._protocols import CommMessage
```
manzt, line 41: Just so I'm understanding, the guarded import here lets us use `CommMessage` in type annotations below (without quotes) but `from __future__ import annotations` turns it into a string for the interpreter?

anywidget/_descriptor.py
```diff
@@ -0,0 +1,615 @@
...
+    if not _is_traitlets_object(obj):
+        return None
+
+    obj_ref = weakref.ref(obj)
+
+    @obj.observe
+    def _on_trait_change(change: dict):
+        name = change["name"]
+        if obj_ref().trait_metadata(name, _TRAITLETS_SYNC_FLAG):
+            send_state({name})
```
manzt, line 589: I guess an alternative here would be to only `observe` to the known traits with `sync=True` metadata (rather than have this condition every change event.

```python
obj.observe(
    lambda change: send_state({ change["name"] }),
    names=list(obj.traits(sync=True).keys()),
)
```

This would of course prevent users from _dynamically_ setting traits. 

```python
counter = CounterWidget()
counter.add_traits(name=traitlets.Unicode().tag(sync=True))
```

So maybe the above is preferred.

anywidget/_descriptor.py
```diff
@@ -0,0 +1,615 @@
...
+
+
+def _get_pydantic_state(obj: pydantic.BaseModel) -> dict:
+    """Get the state of a pydantic BaseModel instance.
+
+    To take advantage of pydantic's support for custom encoders (with json_encoders)
+    we call obj.json() here, and then cast back to a dict (which is what the comm
+    expects.)
+    """
+    return json.loads(obj.json())
```
manzt, line 615: Yeah really a fan of this

manzt: Although kind of a bummer that we need deserialize the JSON into a dict. I guess this is something we could follow up with some special handling in the font-end:

```python
def _get_pydantic_state(obj: pydantic.BaseModel) -> dict:
    return { "_pydantic_json": obj.json() }
```

manzt (approved):
Awesome! Thanks for all the tests too :) Let's get this merged and start thinking about shorthands for composing the descriptor pattern.

anywidget/_descriptor.py
tlambert03: correct.  the most complete (if "dryest") explanation is in the PEP: https://peps.python.org/pep-0563/)

also worth nothing, `_protocols` is actually never imported anywhere (yet) when running anywidget at runtime, since the only import is behind if TYPE_CHECKING.  but we'd need to add typing_extensions to the deps if it was imported outside of TYPE_CHECKING

anywidget/_descriptor.py
```diff
@@ -0,0 +1,615 @@
...
+
+def _connect_traitlets(obj: object, send_state: Callable) -> Callable | None:
+    """Check if an object has a psygnal.SignalGroup, and connect it to send_state.
+
+    Returns
+    -------
+    disconnect : Callable | None
+        A callable that disconnects the psygnal.SignalGroup from send_state, or None
+        if no psygnal.SignalGroup was found.
+    """
```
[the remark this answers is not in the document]
tlambert03: oops :)

anywidget/_descriptor.py
tlambert03: ah good point.  yeah I'm torn.

anywidget/_descriptor.py
tlambert03: yeah, and I guess we could also look at the Comm object itself.  since we're creating it now, we could look at that source and possible modify the final serialization

anywidget/_descriptor.py
manzt: > with a slight preference for only observing a set of names to begin with

Same. tbh I can't really think of a case where dynamically setting state to sync with the front end makes sense. The JS is defined statically, I think it's fair to also make the traits be defined statically.

tlambert03: all set

anywidget/_descriptor.py
tlambert03: let's defer this for a later PR

tlambert03:
(hope you don't mind the merge :) ... onwards)

manzt:
yes! I was just about to ping you. Merge away!

PR #50, opened 2023-02-05
tests: add test coverage with codecov

tlambert03:


tlambert03:
@manzt ... I think you'll need to create an account at codecov to get repos in your user/org working for storing reports:
https://docs.codecov.com/docs

tlambert03:
i don't think you'll need steps 2-4 ... just 1

pyproject.toml
```diff
@@ -21,6 +21,7 @@ dependencies = [
 test = [
     "black[jupyter]",
     "pytest",
+    "pytest-cov",
     "ruff==0.0.238"
 ]
```
manzt, line 26: I use `hatch` CLI locally and these deps are repeated in `[tool.hatch.envs.default]`.  We should add `pytest-cov` there as well I think. 

side note: do you know if it's possible to avoid this duplication? i.e., make `optional-dependencies` dynamic or force hatch to install optional deps in the default envs?

manzt (approved):
Awesome, thank you! I think I set it all up and now just need to merge into `main`.

pyproject.toml
tlambert03: yeah, you can avoid it with `features`:
```py
[tool.hatch.envs.default]
features = ["test"]
```

will do here

manzt: oh, slick!

PR #52, opened 2023-02-05
fix: dynamically set _ipython_display_ for widgets in Google Colab

### LINKED ISSUE #48: bug: widgets don't display with ipython v8.0 in Google Colab
Hard to know if this is something that should be solved upstream in `ipywidgets`, but it is certainly something we can try to resolve for **anywidget** users. I ran into this issue whilst re-implementing the low-level `ipywidgets.Widget` base class which `anywidget.AnyWidget` inherits from.

## Description

Third-party widgets don't work in Google Colab for ipywidgets v8. I've seen various issues related to core ipywidgets not working in Colab, but that has been resolved (read: `ipywidgets.IntSlider` works in v8 but custom widgets do not) and I haven't seen any other solution except "install an older version of ipywidgets".

### ipywidgets v7.7.1 (default in Google Colab)

Everything works appropriately.

### ipywidgets v8.0.1 (manually installed into notebook)

Fails to find third-party widget. No output, obscure error in the browser console.

Screen Shot 2023-02-04 at 3 42 21 PM

## Explanation

I haven't seen this documented anywhere, so I'll do my best to describe the issue to my understanding. Of course Google Colab kernel is closed source, so I can't easily explore the source code to see how objects are displayed like in Jupyter. (read: I dug around in Colab notebook for a couple of hours trying to reverse engineer what the required protocol is.)

To my underestanding, the core issue is a related to a change in the protocols used to display Jupyter Widgets in notebooks from `ipywidgets` v7 to v8.

-  v7 uses the `_ipython_display_` method (which doesn't return anything but as a side effect displays the object with `IPython.display`)
-  v8 uses the `_repr_mimebundle_` method which returns a dictionary of multiple formats keyed by mime-type and then _defers_ to the kernel to display the appropriate mime-type

You can learn more about what each of these methods do in the [IPython docs](https://ipython.readthedocs.io/en/stable/config/integrating.html), but the important caveat is `_ipython_display_` takes the highest priority of all `_repr_*` methods. "If [`_ipython_display_`] is defined, all other display methods are ignored.". It seems that there were [some issues with the `_ipython_display_` protocol](https://github.com/jupyter-widgets/ipywidgets/pulls?q=is%3Apr+sort%3Aupdated-desc+_repr_mimebundle_+is%3Aclosed) for some built-in widgets which act as containers for other widgets (e.g., `ipywidgets.Output`), which lead to the use of `_repr_mimebundle_` in the latest version.

For whatever reason, **Google Colab only supports rendering core `ipywidgets` with the new `_repr_mimebundle_` protocol**, not third-party widgets. However, manually calling `IPython.display` from within `_ipython_display_` works for third-party widgets  seems to work?

## Solution

I was able to get **anywidget** or any other third-party widgets to render in Google Colab by adding a blanket polymorphic `_ipython_display_` method. I.e., the following will work in Google Colab for ipywidgets v7 and v8:

```python
import anywidget
import ipywidgets

class PatchedAnyWidget(anywidget.AnyWidget):
    _esm = "export function render(view) { view.el.innerText = 'Hello, world'; }"

    def _ipython_display(self, **kwargs):
        if hasatter(super(), "_ipython_display_"):
            # ipywidgets v7
            super()._ipython_display_(**kwargs)
        else:
            from IPython.display import display
            
            # ipywidgets v8
            data = self._repr_mimebundle__(**kwargs)
            display(data, raw=True)
```

My main question is whether we should try to patch `_ipython_display_` _only_ in Google Colab (i.e., conditionally with `if "colab" in sys.modules`) or across all environments. I don't fully understand the reasoning from the ipywidgets core team for preferring `_repr_mimebundle_` but importantly if added `_ipython_display_` to `anywidget.AnyWidget` we would revert the changes since the latter protocol takes precedence.

manzt:
Fixes #48

Dynamically adds `_ipython_display_` method to an instance of `anywidget.widget.AnyWidget` if both:

1. detect we are in a Google Colab and 
2. `_ipython_display_` is not already implemented

Importantly, this PR changes the previous behavior of calling `google.colab.outputs.enable_custom_widget_manager()` to within the `anywidget.AnyWidget` constructor. From my experimenting, there seems to be no issue with executing this function each time we create a new instance. ~I think ideally we call this enable function only the _first_ time we detect a widget instance in a colab. Maybe some state that lives on the class?~

EDIT: The enable_custom_widgets_manager is now only excuted once. Originally I had a `functools.cache` init function, but didn't realize that wasn't available in older python 3 versions. There might be a more elegant way to do it than the path I've chosen.

### DIFF
--- a/anywidget/widget.py
+++ b/anywidget/widget.py
@@ -52,13 +52,18 @@ class AnyWidget(ipywidgets.DOMWidget):
 
         self.add_traits(**anywidget_traits)
 
+        # Check if we are in Colab
         if "google.colab.output" in sys.modules:
+
+            # Enable custom widgets manager so that our widgets display in Colab
+            # https://github.com/googlecolab/colabtools/issues/498#issuecomment-998308485
             if not type(self)._enabled_colab_widget_manager:
                 output = sys.modules.get("google.colab.output")
                 output.enable_custom_widget_manager()  # type: ignore
                 type(self)._enabled_colab_widget_manager = True
 
-            # monkey-patch _ipython_display_ for Google Colab
+            # Monkey-patch _ipython_display_ for each instance if missing.
+            # Necessary for Colab to display third-party widget
             # see https://github.com/manzt/anywidget/issues/48
             if not hasattr(self, "_ipython_display_"):
 
--- a/anywidget/widget.py
+++ b/anywidget/widget.py
@@ -52,14 +52,10 @@ class AnyWidget(ipywidgets.DOMWidget):
 
         self.add_traits(**anywidget_traits)
 
-        if "google.colab" in sys.modules:
+        if "google.colab.output" in sys.modules:
             if not type(self)._enabled_colab_widget_manager:
-                import contextlib
-
-                with contextlib.suppress(ImportError):
-                    from google.colab import output  # type: ignore
-
-                    output.enable_custom_widget_manager()
+                output = sys.modules.get("google.colab.output")
+                output.enable_custom_widget_manager()  # type: ignore
                 type(self)._enabled_colab_widget_manager = True
 
             # monkey-patch _ipython_display_ for Google Colab
--- a/anywidget/widget.py
+++ b/anywidget/widget.py
@@ -1,4 +1,5 @@
 import sys
+import functools
 
 import ipywidgets
 import traitlets.traitlets as t
@@ -17,6 +18,16 @@ export function render(view) {
   </p>`;
 }
 """
+
+
+@functools.cache
+def _init_colab():
+    import contextlib
+
+    with contextlib.suppress(ImportError):
+        from google.colab import output  # type: ignore
+
+        output.enable_custom_widget_manager()
 
 
 class AnyWidget(ipywidgets.DOMWidget):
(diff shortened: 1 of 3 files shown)


tests/test_colab.py
```diff
@@ -0,0 +1,63 @@
...
+
+    with mock.patch.dict(
+        sys.modules,
+        get_patch_dict(
+            "google.colab.output",
+            fake_module(enable_custom_widget_manager=enable_custom_widget_manager),
+        ),
+    ):
+        anywidget.AnyWidget()
+        anywidget.AnyWidget()
```
manzt, line 42: Make sure we only call setup function once

anywidget/widget.py
```diff
@@ -48,3 +51,25 @@ def __init__(self, *args, **kwargs) -> None:
         ).tag(sync=True)
 
         self.add_traits(**anywidget_traits)
+
+        # Check if we are in Colab
+        if "google.colab.output" in sys.modules:
+            # Enable custom widgets manager so that our widgets display in Colab
+            # https://github.com/googlecolab/colabtools/issues/498#issuecomment-998308485
+            if not type(self)._enabled_colab_widget_manager:
```
tlambert03, line 59: it's not particularly better... but if you want to ensure that a function only gets called once, you can use `functools.lru_cache`

```python
@lru_cache(maxsize=None)
def _enable_custom_widget_manager():
    # Enable custom widgets manager so that our widgets display in Colab
    # https://github.com/googlecolab/colabtools/issues/498#issuecomment-998308485
    output = sys.modules.get("google.colab.output")
    output.enable_custom_widget_manager()  # type: ignore
```

and then 

```py
if "google.colab.output" in sys.modules:
    _enable_custom_widget_manager()
```

tests/test_colab.py
```diff
@@ -0,0 +1,47 @@
...
+def fake_module(**kwargs: dict):
+    return collections.namedtuple("module", *kwargs.keys())(**kwargs)
+
+
+@pytest.fixture
+def mock_colab():
+    enable_custom_widget_manager = MagicMock()
+    mock_module = fake_module(enable_custom_widget_manager=enable_custom_widget_manager)
+    with patch.dict(sys.modules, {"google.colab.output": mock_module}):
+        yield mock_module
```
tlambert03, line 20: very minor suggestion, these two could be replaced with:

```py
@pytest.fixture
def mock_colab(monkeypatch: pytest.MonkeyPatch):
    mock_module = MagicMock()
    monkeypatch.setitem(sys.modules, "google.colab.output", mock_module)
    return mock_module
```

tlambert03 (approved):
looks good!

anywidget/widget.py
manzt: ah ok... I actually did this with `functools.cache` originally, but then that failed for older versions for Python! I like this better.

tlambert03: yeah, `cache` came later, but `lru_cache` is there for older versions

tests/test_colab.py
manzt: nice

PR #56, opened 2023-02-06
add pre-commit config

tlambert03:
let's discuss this one... this adds a pre-commit config, but ~also adds mypy to the precommit with strict type checking~ (that is moved to #59).

don't necessarily need to go in this direction, just want to open a PR on it before I forget about it.

**update**:
ok, the main thing this PR now does is:
1. update the ruff config to opt in to some more stuff (most importantly, isort, but also pyupgrade and docstyle)
2. adds a pre-commit config.  this does _not_ mean you need to run `pre-commit install` locally.... but if you do, it will be run whenever you try to commit (with ruff, that's really not that bad).  If you ever want to skip the checking you can use `git commit --no-verify`
3. uses pre-commit for ci with `pipx run pre-commit run --all-files`  note that an even "nicer" ci experience can be had by enabling `pre-commit.ci`, but that too is optional. it will a) open PRs to update your pre-commit config when dependencies update and b) automatically fix linting errors when possible in PRs


tlambert03:
pulled out the mypy stuff to another PR (pre-commit and mypy can/should be considered separately)

pyproject.toml
```diff
@@ -77,6 +77,40 @@ fmt = "black ."
...
+    "D100", # Missing docstring in public module
+    "D104", # Missing docstring in public package
+    "D107", # Missing docstring in __init__
+    "D203", # 1 blank line required before class docstring
+    "D212", # Multi-line docstring summary should start at the first line
+    "D213", # Multi-line docstring summary should start at the second line
+    "D401", # Imperative mood
+    "D413", # Missing blank line after last section
+    "D416", # Section name should end with a colon
+]
```
manzt, line 108: Just so I'm aware, should I read any of these as temporary ignores (i.e., eventually we should respect this rule) or not? These are all related to pydocstyle, correct?

pyproject.toml
```diff
@@ -77,6 +77,40 @@ fmt = "black ."
...
+    "D",    # pydocstyle
+    "I",    # isort
+    "UP",   # pyupgrade
+    "S",    # bandit
+    "C4",   # flake8-comprehensions
+    "B",    # flake8-bugbear
+    "A001", # flake8-builtins
+    "TID",  # flake8-tidy-imports
+    "RUF",  # ruff-specific rules
+]
```
manzt, line 97: Generally I'm all for less config, stronger defaults, but it seems to get the most out of ruff requires lots of opt-in. Is that correct? I'm familiary with `E`, `F`, `D`, `I` and now `UP` but have no idea what the rest of these are so I trust you.

manzt (approved):
LGTM! Just some minor clarifying questions.

pyproject.toml
tlambert03: Yes, unfortunately ruff doesn't turn on most of the good stuff by default.  The ones you're familiar with are the big ones (I think isort in particular is important) but I think ruff only includes E and F by default.  RUF is nice since it removed unused type ignores.
The remainder are honesty not that important, I'm actually inclined to remove them here in the interest of less config (they'll rarely hit).

pyproject.toml
tlambert03: docstyle is a little frustrating, due to the multiple docstring conventions (numpy, google, etc...).  This config is for a numpy docs.  It should be noted that ruff now does provide this as

```toml
[tool.ruff.pydocstyle]
convention = "google"  # Accepts: "google", "numpy", or "pep257".
```

unfortunately, that _excludes_ D417 (`Missing argument description in the docstring: {name}`) which is to me the single most important docstring rule (it makes sure your docstrings actually match the parameters in the function).  I also think that there are a lot of annoying/invasive rules in the general D set (things like imperative mood or module docstrings can sometimes make you feel like you're spending more time appeasing the linter than doing useful things).  So this is the config I've been using in all my projects that seems to be the most useful / least invasive.  

That said, I totally agree... it's a lot.  Let me play with it a bit and see if I can get the important bits without the draconian bits in a minimal config

manzt: I see., thanks for the explanation. Too bad there isn't an `extends` for a convention lol. But I agree D417 sounds like the most useful docstring rule, so let's keep it this way.

Don't worry about poking around with it too much, I trust the balance and having the comments is really useful. Maybe we could just add a comment to this like "numpy + D417" so I remember

pyproject.toml
manzt: Ok, in that case let's do `E`, `F`, `D`, `I`, `UP`, and `RUF` :)

manzt:
Feel free to merge this whenever! Happy to keep the `extend-select` as is, or a smaller subset. Trust your opinion.

pyproject.toml
tlambert03: regarding extends for convention, i suspect it will come eventually: https://github.com/charliermarsh/ruff/issues/2606

pyproject.toml
tlambert03: changed

tlambert03:
ok, I trimmed it down a bit and updated the original post to summarize

PR #60, opened 2023-02-07
feat: native HMR

manzt:
Right now "best-in-class" widget development is only enabled when developers setup Vite and use its HMR. However, it's kind of annoying to setup Vite with **anywidget** and get the [two playing together](https://anywidget.dev/en/bundling/). 

This PR is an experiment to enable granular HMR  updates _within_ **anywidget** and be agnostic to bundlers or dev-servers. The trick is that we can send messages to an `AnyModel` any time the front-end code is updated, and then clear existing views and re-render the model. This means we could build adapters on the python side for various JS setups.

https://user-images.githubusercontent.com/24403730/217332349-eec5e4c2-1004-4cef-aff5-995a6f936bae.mov

```python
from psygnal import evented
from dataclasses import dataclass

@evented
@dataclass
class Counter:
    _repr_mimebundle_ = MimeBundleDescriptor()
    _esm: str = "export function render(view) { view.el.innerText = 'oops, nothing to see here' }"
    _css: str = ""
    value: int = 0

counter = Counter()
counter
```

```python
counter._esm = """
  btn.classList.add("counter-button")
    view.model.save_changes();
  });
  view.model.on("change:value", () => { btn.innerHTML = `count is ${count()}` });
  view.el.appendChild(btn);
}
""" # updates the live counter
```

I think this mechanism can be leveraged to setup file-watchers for build assets, and automatically push changes to the frontend during development. My main thought at the moment is whether or not we should use events or send custom messages to the front end. I'm leaning towards the latter but this is a prototype.

### DIFF
--- a/src/widget.js
+++ b/src/widget.js
@@ -92,21 +92,6 @@ async function load_esm(esm) {
 	return widget;
 }
 
-/**
- * @typedef AnyWidgetHMRUpdate
- * @prop {"anywidget:hmr"} type
- * @prop {{ esm?: string, css?: string }} data
- */
-
-/**
- * @param {unknown} msg
- * @return {msg is AnyWidgetHMRUpdate}
- */
-function is_anywidget_update(msg) {
-	if (typeof msg !== "object" || msg == null) return false;
-	return "type" in msg && msg.type === "anywidget:hmr";
-}
-
 /** @param {typeof import("@jupyter-widgets/base")} base */
 export default function (base) {
 	class AnyModel extends base.DOMWidgetModel {
@@ -122,37 +107,34 @@ export default function (base) {
 		initialize(...args) {
 			super.initialize(...args);
 
-			this.on("msg:custom", async (msg) => {
-				if (!is_anywidget_update(msg)) return;
+			this.on("change:_css", () => {
+				load_css(this.get("_css"), this.get("_anywidget_id"));
+			})
 
-				let { esm, css } = msg.data;
+			this.on("change:_esm", async () => {
 
-				if (typeof esm === "string") {
-					let widget = await load_esm(esm);
-					for await (let view of Object.values(this.views ?? {})) {
-						// Unsubscribe from any handlers registered to `view.listenTo`
-						// Sadly we can't just `model.off` because it removes everything,
-						// including handlers not setup by the child view.
-						//
-						// We could override `model.on` with a particular `context` if none is
-						// provided, letting us unsubscribe from only events from views
-						// e.g., `model.off(null, null, anywidgetSymbol)`, but that might
-						// be more trouble than it's worth and this is just a feature for
-						// development.
-						view.stopListening(this);
+				let widget = await load_esm(this.get("_esm"));
 
-						// Clean up all child elements except for the root
-						while (view.el.firstChild) {
-							view.el.removeChild(view.el.firstChild);
-						}
+				for await (let view of Object.values(this.views ?? {})) {
+					// Unsubscribe from any handlers registered to `view.listenTo`
+					// Sadly we can't just `model.off` because it removes everything,
+					// including handlers not setup by the child view.
+					//
+					// We could override `model.on` with a particular `context` if none is
+					// provided, letting us unsubscribe from only events from views
+					// e.g., `model.off(null, null, anywidgetSymbol)`, but that might
+					// be more trouble than it's worth and this is just a feature for
+					// development.
+					view.stopListening(this);
 
-						widget.render(view);
+					// Clean up all child elements except for the root
+					while (view.el.firstChild) {
+						view.el.removeChild(view.el.firstChild);
 					}
+
+					widget.render(view);
 				}
 
-				if (typeof css === "string") {
-					load_css(css, this.get("_anywidget_id"));
-				}
 			});
 		}
 	}
--- a/src/widget.js
+++ b/src/widget.js
@@ -92,6 +92,21 @@ async function load_esm(esm) {
 	return widget;
 }
 
+/**
+ * @typedef AnyWidgetHMRUpdate
+ * @prop {"anywidget:hmr"} type
+ * @prop {{ esm?: string, css?: string }} data
+ */
+
+/**
+ * @param {unknown} msg
+ * @return {msg is AnyWidgetHMRUpdate}
+ */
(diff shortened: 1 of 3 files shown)


manzt:
Thinking we could eventually have an API like:

```python

@anywidget("./index.js", watch=True) # or auto-magically watch if in development install
@evented
@dataclass
def Counter:
    value: int = 0
```

manzt:
poor mans file-watcher :) guessing you have some ideas here, @tlambert03 

https://user-images.githubusercontent.com/24403730/217356900-7b524179-a787-45de-8131-be7c36341ac8.mov



<details>

<summary> code </summary>

```python
import threading
import time
import pathlib

from watchdog.observers import Observer
from watchdog.events import FileSystemEventHandler

class BackgroundUpdater:
    _thread: threading.Thread | None = None
    _should_stop: bool = False
    
    def __init__(self, widget, esm: str | pathlib.Path):
        self._widget = widget
        self._esm = pathlib.Path(esm).absolute()
        
    def stop(self):
        if self._thread is None:
            return
        self._should_stop = True
        self._thread.join()
        self._thread = False
        self._should_stop = False
    
    def start(self):
        if self._thread is not None:
            self.stop()
        
        def target():
            that = self
                    
            class Handler(FileSystemEventHandler):
                def on_modified(self, event):
                    # print(f"{event.event_type=}, {event.src_path=}, {that._esm=}")
                    if event.event_type == "modified" and pathlib.Path(event.src_path) == that._esm:
                        that._widget._esm = that._esm.read_text()
                
            observer = Observer()
            observer.schedule(Handler(), str(self._esm.absolute().parent), recursive = False)
            observer.start()
            
            while not self._should_stop:
                time.sleep(1)
                
            observer.stop()
            observer.join()
                
        self._thread = threading.Thread(target=target, daemon=True)
        self._thread.start()
```
</details>

manzt:
One thought is that if we use custom messages instead of `change` events, we could group comms by `_anywidget_id` to push updates to _all_ active models. https://github.com/jupyter-widgets/ipywidgets/blob/main/packages/schema/messages.md#custom-messages-custom

The `AnyModel` could listen for `msg:custom` with a payload:

```python
comm.send({
  'comm_id': 'u-u-i-d',
  'data': {
    'method': 'custom',
    'content': { 'type': 'anywidget:update', "data": { "esm": new_esm } },
  }
})
```

manzt:
Ok, this PR now adds a `ReprMimeBundle.send_hmr_update` which allows the frontend code to be updated dynamically. 

```python
import pathlib

from psygnal import evented
from dataclasses import dataclass

@evented
@dataclass
class Counter:
    _repr_mimebundle_ = MimeBundleDescriptor()
    value: int = 0

counter = Counter()
counter
```

```python
counter._repr_mimebundle_.send_hmr_update(
    esm=pathlib.Path("./index.js").read_text(),
    css=".counter-button { font-size: 4em; border-radius: 10px; }",
)
```

src/widget.js
```diff
@@ -102,6 +117,44 @@ export default function (base) {
 		static view_name = "AnyView";
 		static view_module = name;
 		static view_module_version = version;
+
+		/** @param {Parameters<InstanceType<base["DOMWidgetModel"]>["initialize"]>} args */
```
manzt, line 121: Ugly hack to make sure we appropriately get the parameter types. Keeping the anywidget src as JS instead of TS makes it so that we don't have to maintain a webpack config on for the `jupyter labextension build` (the default config currently works out of the box).

manzt:
Ok, I think this is ready for review and opens up a world of possibilities for additional APIs. We could probably export an `anywidget.hmr` module for anywidget developers to wrap their widgets during development and automatically push changes for watched files.

manzt:
hmm actually, the more I think about this the ESM/CSS should be a part of the model "state" (other wise we are responsive for syncing custom messages). Switching back to listening to `change:_esm` and `change:_css` events, but keeping the special `send_hmr_update` method.

tlambert03 (approved):
awesome!  I think the only suggestion I have would be to make `_send_hmr_update` private for now.  Ultimately, I think it's something that anywidget should do for the user, as you've shown above with watchdog for instance (based on some watch flag), and it'd be nice if an end-user didn't need to call this (easy to make public in the future too if you change your mind)

tlambert03:
>  here's a minimal file-watcher :) guessing you have some ideas here, @tlambert03

yeah, i love it.  the core of watchdog is little more than a thread that checks `os.stat` on a file and emits an event when it changes.  So I that bit could be done without a new dependency, but this is super cool.  I think if the string passed to `esm` _is_ a local path, this should just be on by default, as you said.

tests/test_descriptor.py
```diff
@@ -77,6 +77,47 @@ def _get_anywidget_state(self):
...
+    )
+
+    foo._repr_mimebundle_._send_hmr_update(css=css)
+    mock_comm.send.assert_called_with(
+        data={"method": "update", "state": {"_css": css}, "buffer_paths": []},
+        buffers=[],
+    )
+
+    foo._repr_mimebundle_._send_hmr_update()
+    mock_comm.send.assert_not_called
```
manzt, line 118: @tlambert03 is this the correct way to assert that the send method doesn't get called? Not deeply familiar with `MagicMock`.

src/widget.js
```diff
@@ -102,6 +102,41 @@ export default function (base) {
...
+		initialize(...args) {
+			super.initialize(...args);
+
+			this.on("change:_css", () => {
+				load_css(this.get("_css"), this.get("_anywidget_id"));
+			})
+
+			this.on("change:_esm", async () => {
```
manzt, line 114: listening for `change` events here instead of custom messages will also make this easier to extend whatever "watcher" we come up with for widgets that derive from the ipywidgets/traitlests based `anywidget.AnyWidget`.

manzt:
>  So I that bit could be done without a new dependency, but this is super cool. I think if the string passed to esm is a local path, this should just be on by default, as you said.

Wow, did not know that! Just googled "file-watcher python", but really nice that we could make a dependency-less helper for this

manzt:
Have you seen https://github.com/samuelcolvin/watchfiles? Its not polling-based like other watchers implementations

tlambert03:
I think I have, but I had forgotten.  That's lovely, use it :)

PR #61, opened 2023-02-08
feat: add `FileContents` emitter

manzt:
_EDITED: click on history to see the prior state of this PR_

This PR adds the `anywidget._file_contents.FileContents` object. `FileContents` watches for file changes in a provided path in a background thread and continuously emits `changed` signals any time the file changes, until the file is deleted (emits a `deleted` event). Connecting `FileContents` with `MimeReprBuilder` or `AnyWidget` allows for #60 to be leveraged during development. Currently this must be done manually, but we should aim to make the creation of `FileContents` hidden.

#### With `traitlets`

```python
import anywidget
import traitlets
from anywidget._file_contents import FileContents

contents = FileContents("index.js")

    _esm = str(contents)

counter = Counter()
@contents.changed.connect(lambda change: counter._esm = change)
counter
```

#### With `MimeReprBuilder`

```python
import anywidget
from anywidget._file_contents import FileContents
from anywidget._descriptor_ import MimeReprBuilder

contents = FileContents("index.js")

@evented
@dataclasses
class Counter:
    _repr_mimebundle_ = MimeReprBuilder(_esm=str(contents)

counter = Counter()
contents.changed.connect(lambda change: counter._send_hmr_update(esm=change))
counter
```


anywidget/_descriptor.py
```diff
@@ -286,6 +294,25 @@ def __init__(
...
+        # read ESM/CSS files if provided and setup watch handlers
+        path = pathlib.Path(self._extra_state[_ESM_KEY])
+        try:
+            if path.is_file():
+                self._extra_state[_ESM_KEY] = path.read_text()
+                if watch:
+                    watcher.watch(path, lambda x: self._send_hmr_update(esm=x))
+        except OSError:
+            ...
+
```
manzt, line 306: I know this is pretty clumsy... just wasn't sure how to handle it

anywidget/_descriptor.py
```diff
@@ -106,6 +109,16 @@ def _comm_for(obj: object) -> Comm:
...
 
 
+def _is_existing_file(x: Any) -> bool:
+    if not isinstance(x, (str, pathlib.Path)):
+        return False
+
+    with contextlib.suppress(OSError):
+        return pathlib.Path(x).is_file()
+
+    return False
```
manzt, line 119: Alternatively, I think this could be:

```python
def _try_read_as_file(x: Any) -> None | tuple[pathlib.Path, str]:
    ...
```

tlambert03:
generally love the idea!  my first thoughts are that

1. i feel like the watching functionality is a bit distributed, the `watch_files` function and the protocol of having a `_watchable` attribute has a slight smell
2. i think the `BackgroundFileWatcher` singleton that can add files (stopping and restarting one thread as needed) is probably a bit overly complicated for now.  Unless you think there will be tens-hundreds of files being watched, I think i'd just have an object per file being watched.
3. The file watcher is also recreating the observer pattern by managing all of the `FileChangeHandler` objects.  You could consider using psygnal for that, and lose a lot of code.

I'm thinking something like this:

```py
from psygnal import Signal
from watchfiles import watch, Change


class FileContents:
    """Object that watches a file for changes and emits a signal when it changes.
    
    Calling `str(obj)` on this object will always return the current contents of the
    file (as long as the thread is running).
    """

    changed = Signal(str)
    deleted = Signal()

    def __init__(self, path: str | pathlib.Path, start_thread: bool = True):
        self._path = pathlib.Path(path).expanduser().absolute()
        self._contents: str | None = None  # cached contents, cleared on change
        self._stop_event = threading.Event()
        self._background_thread: threading.Thread | None = None
        if start_thread:
            self.watch_in_thread()

    def watch_in_thread(self) -> None:
        if self._background_thread is not None:
            return
        self._stop_event.clear()
        self._background_thread = threading.Thread(target=self.watch, daemon=True)
        self._background_thread.start()

    def stop_thread(self) -> None:
        if self._background_thread is None:
            return
        self._stop_event.set()
        self._background_thread.join()
        self._background_thread = None

    def watch(self) -> Iterator[tuple[int, str]]:
        for changes in watch(self._path, stop_event=self._stop_event):
            for change, path in changes:
                if change == Change.deleted:
                    self.deleted.emit()
                    return
                if change == Change.modified:
                    self._contents = None
                    self.changed.emit(str(self))
                    yield (change, path)
                    break

    def __str__(self) -> str:
        if self._contents is None:
            self._contents = self._path.read_text()
        return self._contents
```

Then, you can always just call `str(obj)` and get the latest contents.  So, somewhere else in the code, it would be something like:

```python
esm = ...  # some string
if Path(esm).is_file():
    contents = FileContents()
    @contents.changed.connect
    def _on_change(new_contents):
                self._extra_state['_esm'] = contents
                self.send_state({'_esm'})
```

I know a little bit more needs to be done to work for both  esm and css... but i think id keep the handler logic on the respective objects, rather than having the watch_files function that checks what kind of object it's working with.

(side note: I also think it'd be nice to unify those two objects... the `AnyWidget | AnywidgetProtocol` should be temporary)

manzt:
Oooo I really like the `FileContents` :)

> Unless you think there will be tens-hundreds of files being watched, I think i'd just have an object per file being watched.

I agree the global watcher is probably too complicated and this would be more simple. However, I need to investigate whether `watchfiles` allows for multiple listeners to the same directory. If you try to listen to the same directory with `watchdog`, it throws an error. Also it would be nice if we could create a single `FileContents` per file, and then connect children to that `FileContents` instance (similar to `COMMS`):

```python

esm = ...
if Path(esm).is_file():
    contents = _file_contents_for(esm)
    @contents.changed.connect
    def _on_change(new_contents):
        ...
```

What do you think?

manzt:
Lol, here's a crazy idea why not just have widget authors be explicit about this feature..

```python
import anywidget
from anywidget._file_contents import FileContents
import traitlets

    _esm = FileContents("index.js")
    _css = FileContents("index.css")

counter = Counter()
counter
```

manzt:
also I think with this PR we can deprecate the `ReprMimeBundle._send_hmr_update` API and restrict the ability to send HMR updates to `FileContents`.

PR #62, opened 2023-02-09
feat: add `FileContents` to read/watch files

manzt:
Closes #61

This PR adds the `anywidget._file_contents.FileContents` object. `FileContents` watches for file changes in a provided path in a background thread and continuously emits `changed` signals any time the file changes, until the file is deleted (emits a `deleted` event). Connecting `FileContents` with `MimeReprBuilder` or `AnyWidget` allows for #60 to be leveraged during development. Currently this must be done manually, but we should aim to make the creation of `FileContents` hidden.

#### With `traitlets`

```python
import anywidget
import traitlets
from anywidget._file_contents import FileContents

contents = FileContents("index.js")

    _esm = str(contents)

counter = Counter()
contents.changed.connect(lambda change: counter._esm = change)
counter
```

#### With `MimeReprBuilder`

```python
import anywidget
from anywidget._file_contents import FileContents
from anywidget._descriptor_ import MimeReprBuilder

contents = FileContents("index.js")

@evented
@dataclasses
class Counter:
    _repr_mimebundle_ = MimeReprBuilder(_esm=str(contents)

counter = Counter()
contents.changed.connect(lambda change: counter._send_hmr_update(esm=change))
counter
```

### DIFF
--- a/anywidget/_file_contents.py
+++ b/anywidget/_file_contents.py
@@ -1,0 +1,84 @@
+from __future__ import annotations
+from collections import deque
+
+import pathlib
+import threading
+from typing import Iterator
+
+from psygnal import Signal
+
+
+class FileContents:
+    """Object that watches for file changes and emits a signal when it changes.
+
+    Calling `str(obj)` on this object will always return the current contents of the
+    file (as long as the thread is running).
+
+    Parameters
+    ----------
+    path : str | pathlib.Path
+        The file to read and watch for content changes
+    start_thread : bool, optional
+        Whether to start watching for changes in a separate thread (default: `True`)
+    """
+
+    changed = Signal(str)
+    deleted = Signal()
+
+    def __init__(self, path: str | pathlib.Path, start_thread: bool = True):
+        self._path = pathlib.Path(path).expanduser().absolute()
+        if not self._path.is_file():
+            raise ValueError("File does not exist: {self._path}")
+        self._contents: str | None = None  # cached contents, cleared on change
+        self._stop_event = threading.Event()
+        self._background_thread: threading.Thread | None = None
+        if start_thread:
+            self.watch_in_thread()
+
+    def watch_in_thread(self) -> None:
+        """Watch for file changes (and emitting signals) from a separate thread."""
+        if self._background_thread is not None:
+            return
+        self._stop_event.clear()
+        self._background_thread = threading.Thread(
+            target=lambda: deque(self.watch(), maxlen=0),
+            daemon=True,
+        )
+        self._background_thread.start()
+
+    def stop_thread(self) -> None:
+        """Stops an actively running thread if it exists."""
+        if self._background_thread is None:
+            return
+        self._stop_event.set()
+        self._background_thread.join()
+        self._background_thread = None
+
+    def watch(self) -> Iterator[tuple[int, str]]:
+        """Watch for file changes and emit changed/deleted signal events.
+
+        Blocks indefinitely.
+
+        Returns
+        -------
+        changes : Iterator[tuple[int, str]]
+            An iterator that yields any time the file changes until the file is deleted.
+        """
+        from watchfiles import Change, watch
+
+        for changes in watch(self._path, stop_event=self._stop_event):
+            for change, path in changes:
+                if change == Change.deleted:
+                    self.deleted.emit()
+                    return
+                # Only getting Change.added events on macOS so we listen for either
+                if change == Change.modified or change == Change.added:
+                    self._contents = None
+                    self.changed.emit(str(self))
+                    yield (change, path)
+                    break
+
+    def __str__(self) -> str:
+        if self._contents is None:
+            self._contents = self._path.read_text()
+        return self._contents
--- a/anywidget/_file_contents.py
+++ b/anywidget/_file_contents.py
@@ -41,6 +41,9 @@ class FileContents:
             return
         self._stop_event.clear()
         self._background_thread = threading.Thread(
+            # self.watch() returns a generator, so the thread would exit
+            # immediately if we passed `target=self.watch`. The `deque`
+            # forces the thread to run until it exhausts our generator.
             target=lambda: deque(self.watch(), maxlen=0),
             daemon=True,
         )
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,6 +26,7 @@ test = [
     "pytest",
     "pytest-cov",
     "ruff==0.0.238",
+    "watchfiles",
 ]
 
 [project.urls]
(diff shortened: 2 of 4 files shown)


manzt:
Lol, not quite sure why the coverage is still messaged up. Anyways, I think this is probably ready for a review now. Like I said, still thinking about the right API to hide the `FileContents` for anywidget.

anywidget/_file_contents.py
```diff
@@ -0,0 +1,84 @@
...
+        if start_thread:
+            self.watch_in_thread()
+
+    def watch_in_thread(self) -> None:
+        """Watch for file changes (and emitting signals) from a separate thread."""
+        if self._background_thread is not None:
+            return
+        self._stop_event.clear()
+        self._background_thread = threading.Thread(
+            target=lambda: deque(self.watch(), maxlen=0),
```
tlambert03, line 44: tell me about this line.  if the lambda and deque are added to make it work (as opposed to just using `target=self.watch`) then we could also consider removing the yield from the `watch` method (which turns it into a generator function).

tests/test_file_contents.py
```diff
@@ -0,0 +1,162 @@
...
+
+    CONTENTS = "hello, world"
+    path = tmp_path / "foo.txt"
+    path.write_text(CONTENTS)
+
+    contents = FileContents(path, start_thread=False)
+
+    assert str(contents) == CONTENTS
+    assert contents._background_thread is None
+    mock.assert_not_called
```
tlambert03, line 28:
```suggestion
def test_file_contents_no_watch(tmp_path: pathlib.Path):
    """Test __str__ reads file contents and does not start a thread"""
    CONTENTS = "hello, world"
    path = tmp_path / "foo.txt"
    path.write_text(CONTENTS)

    with patch.object(watchfiles, "watch") as mock:
        contents = FileContents(path, start_thread=False)
        assert str(contents) == CONTENTS
        assert contents._background_thread is None
    mock.assert_not_called()
```

this is the same exact thing... but `with patch...` is a nice way to accomplish this with less code (add `from unittest.mock import patch` at the top).  pytest's monkeypatch fixture uses it under the hood, and in this particular case, it's a little more direct.  but, absolutely the same thing :)

if you like it, it could be used below as well, but also fine to stick with the current code

anywidget/_file_contents.py
tlambert03: maybe nevermind... I see that you used `obj.watch` in tests...  which *is* nice if you want to retain the non-threaded behavior

manzt: That's exactly it. `target=self.watch` just has the thread execute `self.watch()` which immediately returns a generator, so the the thread would terminate without entering an indefinite loop.

I saw that `deque(generator, maxlen=0)` is an [efficient way to consume a iterable](https://stackoverflow.com/questions/47456631/simpler-way-to-run-a-generator-function-without-caring-about-items) so I added that to have the thread exhaust the generator. You're right that the we could remove the `yield` keyword (changing `watch` from a generator to a regular function) and remove the need for this. I'm fine either way.

tlambert03: nah, I later I realized that you do use it in tests... so it's nice to keep the functionality. let's stick with it.

tlambert03 (commented):
(still looking)

anywidget/_file_contents.py
manzt: Sounds good. I'll add a comment

tests/test_file_contents.py
manzt: love it. Thanks!

tests/test_file_contents.py
```diff
@@ -0,0 +1,157 @@
...
+    monkeypatch.setattr(watchfiles, "watch", mock_watch)
+
+    contents = FileContents(path, start_thread=False)
+    mock = Mock()
+    contents.deleted.connect(mock)
+
+    total = sum(1 for _ in contents.watch())
+    assert total == 0
+    mock_watch.assert_called_with(path, stop_event=contents._stop_event)
+    assert mock.called
```
tlambert03, line 50: another optional alternative here:

```suggestion
def test_file_contents_deleted(tmp_path: pathlib.Path):
    """Test deleting a file emits a deleted signal and stops the watcher"""
    CONTENTS = "hello, world"
    path = tmp_path / "foo.txt"
    path.write_text(CONTENTS)

    contents = FileContents(path)

    mock = Mock()
    contents.deleted.connect(mock)
    path.unlink()
    assert contents._background_thread
    contents._background_thread.join()
    assert not contents._background_thread.is_alive()
    mock.assert_called()
```

this just lets watcher.watch do it's thing with an actual delete event (no mock)

manzt: oh right, this is great!

tlambert03: as a side note here... a handy thing in `unittest`, if you ever _want_ to let an object do its regular job, but also assert that it was indeed called, you can use the `wraps` argument.  so it would be something like:


```python
    with patch.object(watchfiles, "watch", wraps=watchfiles.watch) as mock_watch:
        contents = FileContents(path)
        mock_watch.assert_called_once_with(contents._path, stop_event=contents._stop_event)
```

manzt: good to know. hmm this test is failing now. will poke around a bit later. I think this is related to an issue I was running into earlier where it took a little while to "start" watching so the unlink wasn't observed.

tlambert03: hmm, passing locally... lemme check 3.8

pyproject.toml
```diff
@@ -26,6 +26,7 @@ test = [
     "pytest",
     "pytest-cov",
     "ruff==0.0.238",
+    "watchfiles",
```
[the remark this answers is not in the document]
manzt: I've seen some "%(dev)" syntax before in `setup.cfg` for linking different extras_require, is that supported here?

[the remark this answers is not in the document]
tlambert03: unfortunately no... that `ini`-style include syntax isn't supported in toml, so you either need to add it to both `test` and `dev`... or do `pip install -e .[test,dev]` on ci when you install things for tests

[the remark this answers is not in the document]
manzt: Ah, ok sounds good. I'll just update CI then. Thanks!

tests/test_file_contents.py
tlambert03: arg... you could try this:

```python
    contents = FileContents(path)
    while not contents._background_thread.is_alive():
        time.sleep(0.1)
```


or just go back to before :/

tlambert03 (approved):
this is great.  thanks for writing all the tests.

merge when ready (s'ok if you either want to revert my suggestion, or use the `sleep` approach)

manzt:
Thanks for the review! I tried the `sleep` but it failed in CI, so I'm reverting :) Learned a lot about python unit testing with this PR lol. Love it. thanks!

PR #64, opened 2023-02-11
add support for evented msgspec.Struct objects

tlambert03:
msgspec is a very promising fast serialization library (kinda like pydantic but with more of a focus on the ser/des ... and less of a focus on runtime casting of python types).  It's a good candidate for something anywidget could use as the "default" base class (though, it's somewhat restrictive too, so maybe not).

This pr detects when we're working with evented msgspec.Structs, just as with pydantic, traitlets, etc.

Tests might fail for now cause psygnal itself is also just adding support for msgspec


anywidget/_descriptor.py
```diff
@@ -635,3 +643,17 @@ def _get_pydantic_state(obj: pydantic.BaseModel) -> dict:
...
+    """Return `True` if an object is an instance of msgspec.Struct."""
+    msgspec = sys.modules.get("msgspec")
+    return isinstance(obj, msgspec.Struct) if msgspec is not None else False
+
+
+def _get_msgspec_state(obj: msgspec.Struct) -> dict:
+    """Get the state of a msgspec.Struct instance."""
+    import msgspec
+
+    return msgspec.json.decode(msgspec.json.encode(obj))
```
tlambert03, line 659: particularly painful here...  but we'll address this in a followup

manzt:
whoa interesting, never heard of msgspec before! re-encoding/decoding (e.g., `msgspec.json.decode(msgspec.json.encode(obj))`, I think we could easily address this with a custom deserializer on the front-end:

```python
def _get_msgspec_state(obj: msgspec.Struct):
    """Get the state of a msgspec.Struct instance."""
    import msgspec

    return { "_anywidget_json": msgspec.json.encode(obj) }
```

anywidget/_descriptor.py
jcrist: It's not clear to me if/why this is necessary, but if it is it'd be faster to do `msgspec.to_builtins(obj)` instead. This should have the same end result, but avoid encoding then decoding.

tlambert03: Thanks @jcrist 

The reason this is here (temporarily, as a hack, sadly, for now)... is that we don't yet override the actual Comm object's serialization event.  Comm comes from another library, and expects a dict object and serializes with the stdlib json.  And since json doesn't provide an easy way to define serialization of custom classes _on the class_ (one must customize the serializer itself), here we just try to make sure that the dict will be "plain enough" for json.

This pattern here mostly came from the pydantic pattern used above, and it was earmarked here for later investigation.  

But thank you for pointing out to builtins!  That's great

manzt:
@tlambert03, anything I can do to help push this forward? Would be happy to have support for msgspec!

tlambert03:
will clean up and ping you in a couple days.  just getting back from a two week course and unburying myself :)

tlambert03:
ready for review.  Note the fail on py3.8 was due to an issue with the most recent IPython release: https://github.com/ipython/ipython/issues/14053

hence the workaround in pyproject

manzt (approved):
@tlambert03 totally missed this, I'm so sorry for the delayed review! Thanks so much for the contribution :)

jcrist:
Nice! Mind if I add `anywidget` to the "Used by" section of the msgspec docs? https://jcristharif.com/msgspec/#used-by

manzt:
That would be great, thanks! Logo here [here](https://raw.githubusercontent.com/manzt/anywidget/main/docs/public/favicon.svg) if you need!

jcrist:
Thanks, done here: https://github.com/jcrist/msgspec/pull/421

PR #75, opened 2023-02-17
feat: inject colab meta in `_repr_mimebundle_`

manzt:
Implements the more robust solution for supporting colab outlined in #63, but also for `ReprMimeBundle`.


tlambert03 (approved):
nice!

PR #79, opened 2023-03-01
feat: auto infer `FileContents` for valid file paths

manzt:
This PR adds a useful feature to auto-create `FileContents` objects from `_esm` or `_css` attributes on anywidget subclasses.

```python
import anywidget
import traitlets

    _esm = "index.js"
    value = traitlets.Int().tag(sync=True)

Counter()
```

A `FileContents` object is only created if the provided `str` or `pathlib.Path` points to an existing file. In addition, the object is set to "watch mode" (spinning up a background thread) if the path is detected outside of `site-packages` (i.e., likely a development install).

EDIT:

This PR now also implements supports detection in the descriptor:

```python
from psygnal import evented
from dataclasses import dataclass

@evented
@dataclass
class Counter:
    _repr_mimebundle_ = MimeBundleDescriptor(_esm="index.js")
    count: int = 0

Counter()
```

### DIFF
--- a/anywidget/_descriptor.py
+++ b/anywidget/_descriptor.py
@@ -25,9 +25,18 @@ import weakref
 from dataclasses import asdict, is_dataclass
 from typing import TYPE_CHECKING, Any, Callable, Iterable, Sequence, cast, overload
 
-from ._util import get_repr_metadata, put_buffers, remove_buffers
+from ._file_contents import FileContents
+from ._util import (
+    _ANYWIDGET_ID_KEY,
+    _DEFAULT_ESM,
+    _ESM_KEY,
+    _WIDGET_MIME_TYPE,
+    get_repr_metadata,
+    put_buffers,
+    remove_buffers,
+    try_file_contents,
+)
 from ._version import __version__
-from .widget import DEFAULT_ESM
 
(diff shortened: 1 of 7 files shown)


manzt:
hmm not quite sure what the pytest errors are... The unit tests look to be passing

anywidget/_descriptor.py
```diff
@@ -423,27 +444,6 @@ def unsync_object_with_view(self) -> None:
...
-
-        if esm is not None:
-            update["_esm"] = esm
-
-        if css is not None:
-            update["_css"] = css
-
-        self._extra_state.update(update)
-        self.send_state(update.keys())
-
```
manzt, line 446: This method is really essentially replaced with the new functionality.

manzt:
Up next, I think a `widget` decorator based on the descriptor :) 

```python
from anywidget import widget
from psygnal import evented
from dataclasses import dataclass


@widget(esm="index.js")
@evented
@dataclass
class Counter:
    count: int = 0
```

manzt:
@tlambert03 gentle nudge if you have the time! otherwise no worries :)

tlambert03 (approved):
won't have time this week to pull and play with it, but everything looks pretty good!  I'd say go for it

PR #95, opened 2023-03-25
docs: Add another minimal example

kolibril13:
Another notebook as a template.
Maybe this can even be linked in the "Tips for beginners" section.
https://anywidget.dev/en/jupyter-widgets-the-good-parts/


manzt (commented):
Thanks for the contribution!! I like the idea of adding a notebook with a tutorial showcasing code in separate files. However, one challenging thing is that the notebook is now less portable because it requires `index.js` to be adjacent.

For example, if you open your notebook in colab with the following link:

http://colab.research.google.com/github/kolibril13/anywidget/blob/main/examples/minimal_widget/minimal_example.ipynb

it doesn't copy over the JS file.

I wonder if the notebook could _write_ the JS (from the notebook) and with some explanation of what's going on.

"widgets can either have inlined ESM, <example>, or ..."

```python
with open("index.js") as f:
    f.write("""
    console.log("Hello World 2");
}
""")
```

kolibril13:
good idea!
I just tweaked the example, and it now works in Google Colab! 👍🏼  
and I have used the
```
path = pathlib.Path.cwd() / "index.js"
if not path.exists():
    path.write_text(
```
syntax.
I've added `if not path.exists()` because without it, users might experience an unpleasant surprise when they tweak their index.js file and then accidentally overwrite that file when they re-run the notebook.

manzt:
Looks great, thanks for the contribution!

PR #107, opened 2023-04-13
fix: flexible resolution of @jupyter-widgets/base

### LINKED ISSUE #100: anywidget not working with Panel + ipywidgets-bokeh: Class null not found in module @jupyter-widgets/base@2.0.0
Hi

I am a user of and contributor to [HoloViz Panel](https://panel.holoviz.org). One of the reasons I prefer Panel for my data apps is that it works with ipywidgets. I would like Panel and Anywidget to work together to enable users to easily develop powerful data apps using Anywidget and derived widgets like [mapwidget](https://github.com/giswqs/mapwidget).

Unfortunately it currently does not work.

I don't know which python package causes the issue. So I've cross posted here https://github.com/bokeh/ipywidgets_bokeh/issues/89.

## Reproducible Example

```bash
pip install panel==0.14.4 ipywidgets-bokeh==1.3.0 anywidget==0.2.0
```

**script.py**

```python
import anywidget
import traitlets

import panel as pn

pn.extension("ipywidgets")

class CounterWidget(anywidget.AnyWidget):
    _esm = """
      let getCount = () => view.model.get("count");
      let button = document.createElement("button");
      button.classList.add("counter-button");
      button.innerHTML = `count is ${getCount()}`;
      button.addEventListener("click", () => {
        view.model.set("count", getCount() + 1);
        view.model.save_changes();
      });
      view.model.on("change:count", () => {
        button.innerHTML = `count is ${getCount()}`;
      });
      view.el.appendChild(button);
    }
    """
    _css="""
    .counter-button { background-color: #ea580c; }
    .counter-button:hover { background-color: #9a3412; }
    """
    count = traitlets.Int(0).tag(sync=True)

counter = CounterWidget()
pn.panel(counter).servable()
```

```bash
panel serve script.py
```

Open http://localhost:5006/script and see the exception in the browser console.

image

```bash
Class null not found in module @jupyter-widgets/base@2.0.0
```

Any help would be appreciated. Thanks.

manzt:
Fixes #100

### DIFF
--- a/.changeset/large-bulldogs-matter.md
+++ b/.changeset/large-bulldogs-matter.md
@@ -1,0 +1,5 @@
+---
+"anywidget": patch
+---
+
+fix: allow more flexible semver resolution for `@jupyter-widgets/base`
--- a/packages/anywidget/package.json
+++ b/packages/anywidget/package.json
@@ -26,7 +26,7 @@
 		"build": "node build.mjs"
 	},
 	"dependencies": {
-		"@jupyter-widgets/base": "^6.0.2",
+		"@jupyter-widgets/base": "^2 || ^3 || ^4 || ^5 || ^6",
 		"@anywidget/types": "workspace:*",
 		"@anywidget/vite": "workspace:*"
 	},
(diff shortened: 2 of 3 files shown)


MarcSkovMadsen:
This great. Thanks. How/ when can I test it works for me also?

Issue #104, opened 2023-04-13
Log Error When ESM Import Fails


rmorshea:
I had a pretty tough time debugging an issue that resulted from some bad code that was generated during a build step and which got executed at "import-time" rather than when rendering. This was difficult to figure out because ipywidgets simply reports something like `failed to render model id <the-id>` without logging the original error. This could be mitigated if anywidget reported errors that arose [when importing the ESM module](https://github.com/manzt/anywidget/blob/c74e0518ded7f9a6f4a45d9c57390370da0d5e40/packages/anywidget/src/widget.js#L92).

manzt:
Oof, ya we definitely could improve this. Thanks for the PR. I hadn't considered the unhandled exception for a failed import. Generally there is an issue with obfuscation of the source code because we need to create a blob URL. Still trying to think of ways to improve the "debuggability" of this.

rmorshea:
Thanks for getting that over the finish line!

PR #113, opened 2023-04-20
Fix JS variable scope issue

keller-mark:
There is currently a variable scoping issue in v0.2.2 resulting in the following JS error:

```
Uncaught (in promise) ReferenceError: widget is not defined
```

https://observablehq.com/d/d283036191b4763e


manzt (approved):
Ah, this isn't great. Unfortunately missed this in #104. Will merge and cut a release. Need to figure out an easier way to test the front-end code.

Issue #94, opened 2023-03-25
ModuleNotFoundError: No module named 'watchfiles'


kolibril13:
Following the tutorial from https://anywidget.dev/en/jupyter-widgets-the-good-parts/ (installed with `pip install anywidget`
) will lead to an error because not all dependencies are installed:
`ModuleNotFoundError: No module named 'watchfiles'`
This can be easily fixed:
<img width="572" alt="image" src="https://user-images.githubusercontent.com/44469195/227714861-6d9048a9-379e-40a1-ac69-b66db5ce9c22.png">

Maybe one can add the watchfiles to the pyproject.toml dependency file?
Or one can mention in the https://anywidget.dev/en/getting-started/ page, that one should install `pip install --upgrade "anywidget[dev]"
`

Thanks for this great package!

manzt:
Updated the doc! I have chosen not to include `watchfiles` as a strict dependency because it's only utilized when using the live-reload behavior during development (read: only used by **anywidget** developers when creating a widget, not downstream users of the widget).

kolibril13:
nice, thank you!
When I use the live-reload behavior with `watchfiles` during development, do I have to add the `read_text()` method before shipping a package to pypi?
` _esm = pathlib.Path("index.js").read_text()`
Or would I add the `watchfiles` package to the dependencies for my anywidget project?

PR #116, opened 2023-05-02
fix: add watchfiles as dependency

manzt:
Right now, we raise an exception if one starts developing a widget locally but doesn't have `watchfiles` installed.

Requiring developers to remember to install optional dev dependencies (i.e.,`pip install "anywidget[dev]"`) is kind of annoying, and it's a pretty minimal dependency. This PR adds `watchfiles` as a direct dependency so developers can use `pip install anywidget` and get started without any hiccups.

As an alternative, we could add a warning rather rather than raising an exception, notifying the developer that "watchfiles" must be installed to enable anywidget's live development features, but still load the file the first time. I think adding the dependency is more simple solution, but am open to alternatives.

### DIFF
--- a/.changeset/great-trainers-fetch.md
+++ b/.changeset/great-trainers-fetch.md
@@ -1,0 +1,5 @@
+---
+"anywidget": patch
+---
+
+fix: add `watchfiles` as a directy dependency
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Learn more in
 may be installed with `pip`:
 
 ```bash
-pip install "anywidget[dev]"
+pip install anywidget
 ```
 
 It is also available on
--- a/docs/src/pages/blog/anywidget-02.md
+++ b/docs/src/pages/blog/anywidget-02.md
@@ -13,7 +13,7 @@ _TL;DR: **anywidget** v0.2 brings modern web development to Jupyter. You can now
 use a **file path** to enable **anywidget**'s integrated Hot Module Replacement (HMR):_
 
 ```sh
-pip install --upgrade "anywidget[dev]"
+pip install --upgrade anywidget
 ```
 
 ```python
@@ -193,7 +193,7 @@ to allow for the special JSX syntax (e.g., `<App />`).
 To start using **anywidget** v0.2, upgrade your package using pip:
 
 ```sh
-pip install --upgrade "anywidget[dev]"
+pip install --upgrade anywidget
 ```
 
 I encourage you to explore the new features in **anywidget** v0.2 and experience the
--- a/docs/src/pages/en/getting-started.mdx
+++ b/docs/src/pages/en/getting-started.mdx
@@ -28,7 +28,7 @@ ecosystem, allowing your widget to run automatically in **Jupyter Notebooks**,
 ## Example
 
 ```
-pip install "anywidget[dev]"
+pip install anywidget
 ```
 
 ```python
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,6 +17,7 @@ dependencies = [
     "importlib-metadata ; python_version < '3.8'",
     "typing-extensions>=4.2.0",
     "psygnal>=0.8.1",
+    "watchfiles>=0.18.0",
 ]
 
 [project.optional-dependencies]
@@ -28,7 +29,6 @@ test = [
     "ruff",
 ]
 dev = [
-    "watchfiles",
     "ipykernel",
 ]
 


kolibril13:
Hey @manzt,
I have a few projects with JupyterLite demos that are based on ipyreact, which itself is based on anywidget.
Unfortunately, watchfiles is not available in JupyterLite, therefore all JupyterLite examples are not working anymore with the latest ipyreact version which @maartenbreddels  just released.

when I try to run
%pip install watchfiles
in JupyterLite, I get:
```
ValueError: Can't find a pure Python 3 wheel for 'watchfiles'.
See: https://pyodide.org/en/stable/usage/faq.html#micropip-can-t-find-a-pure-python-wheel
```

A quick workaround fix for now is
`%pip install -q ipyreact anywidget<0.2.4`

Here are two ideas to approach this issue:
1. Remove the watchfiles dependency again
2. Writing an issue in the JupyterLite repo and asking if there is a way to support watchfiles.

And another question in this regard:

When I use the live-reload behavior with `watchfiles` during development, do I have to add the `read_text()` method before shipping a package to pypi?
E.g.
` _esm = pathlib.Path("index.js").read_text()`
Or would I add the `watchfiles` package to the dependencies for my anywidget project?

manzt:
Thanks for opening this issue. I’d made the change since you (https://github.com/manzt/anywidget/issues/94) and others had some friction with developing using anywidget initially. I wasn’t aware of the JupyterLite incompatibility.

I wonder if rather than raising an error like before, we could just have a warning if watchfiles isn’t installed, notifying the developer they need to install it if they want the file to reload automatically.

> When I use the live-reload behavior with watchfiles during development, do I have to add the read_text() method before shipping a package to pypi?

no, watchfiles is only for _reloading_ the file if the contents change. it triggers anywidget to call read_text() internally. Reloading is a development-only feature. When someone pip-installs your widget, the file contents are loaded once.

manzt:
Thanks for opening this issue. I’d made the change since you (https://github.com/manzt/anywidget/issues/94) and others had some friction with developing using anywidget initially. I wasn’t aware of the JupyterLite incompatibility.

I wonder if rather than raising an error like before, we could just have a warning if watchfiles isn’t installed, notifying the developer they need to install it if they want the file to reload automatically.

> When I use the live-reload behavior with watchfiles during development, do I have to add the read_text() method before shipping a package to pypi?

no, watchfiles is only for _reloading_ the file if the contents change. it triggers anywidget to call read_text() internally. Reloading is a development-only feature. When someone pip-installs your widget, the file contents are loaded once.

tlambert03:
yeah, I noticed the recent requirement addition :) Even if it feels like a minor thing, I feel like if it _can_ be optional, it should be optional.  And since watchfiles is only necessary for development, I think you should remove it and just solve things like #94 using documentation, or an inline comment somewhere.

kolibril13:
> no, watchfiles is only for reloading the file if the contents change.

Amazing!
As users don't need it, I also think watchfiles should not be shipped by default.
This is also the easiest solution for the JupyterLite issue.

>  could just have a warning if watchfiles isn’t installed, notifying the developer they need to install it if they want the file to reload automatically

That sounds great!

tlambert03:
> we could just have a warning if watchfiles isn’t installed, notifying the developer they need to install it if they want the file to reload automatically.

yeah I like that too!

nvictus:
I suppose it could also be shipped via `optional-dependencies` with something like `pip install anywidget[complete]` or `anywidget[developer]`

manzt:
@nvictus, that's what we did before 🫣

```
pip install "anywidget[dev]"
```

nvictus:
Ah, right. I noticed it but wasn't sure if `dev` was intended for _development_ on anywidget, or _developers_ of `AnyWidget`s... Some ambiguity there.

maartenbreddels:
awesome, great that you responded so quickly @manzt !

psychemedia:
I am currently trying to test this with the `xeus-python` jupyter kernel, and with `anywidget=0.9.13` in the `environment.yml` file used to build that kernel, I am getting the watchfiles error?

```text
Channels:
 - https://repo.mamba.pm/emscripten-forge
 - conda-forge
 - defaults
Platform: emscripten-wasm32
Collecting package metadata (repodata.json): ...working... done
Solving environment: ...working... failed
LibMambaUnsatisfiableError: Encountered problems while solving:
  - nothing provides requested watchfiles
  - nothing provides watchfiles needed by anywidget-0.9.13-pyhd8ed1ab_0
Could not solve for environment specs
The following packages are incompatible
├─ anywidget 0.9.13  is not installable because it requires
│  └─ watchfiles, which does not exist (perhaps a missing channel);
└─ watchfiles does not exist (perhaps a typo or a missing channel).
```

PR #126, opened 2023-06-01
feat: add experimental module with decorator

manzt:
Adds the `anywidget.experimental` model with a simple decorator implementation.

```python
import dataclasses
import psygnal

from anywidget.experimental import widget

@widget(esm="index.js")
@psygnal.evented
@dataclasses.dataclass
class Counter:
    value: int = 0
```

```python
import msgspec
import psygnal

from anywidget.experimental import widget

@widget(esm="index.js")
@psygnal.evented
class Counter(msgspec.Struct, weakref=True):
    value: int = 0
```

```python
from psygnal import EventedModel # pydantic

from anywidget.experimental import widget

@widget(esm="index.js")
class Counter(EventedModel):
    value: int = 0
```

In a follow-up PR, I'd like to automatically decorate a class with a model and emitter if undecorated. E.g.,

```python
from anywidget.experimental import widget

@widget(esm="index.js")
class Counter:
  value: int = 0
```

TODO: add unit test

### DIFF
--- a/.changeset/pretty-rockets-behave.md
+++ b/.changeset/pretty-rockets-behave.md
@@ -1,0 +1,18 @@
+---
+"anywidget": patch
+---
+
+feat: Add `anywidget.experimental` with simple decorator
+
+```python
+import dataclasses
+import psygnal
+
+from anywidget.experimental import widget
+
+@widget(esm="index.js")
+@psygnal.evented
+@dataclasses.dataclass
+class Counter:
+    value: int = 0
+```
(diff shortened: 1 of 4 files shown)


manzt:
@tlambert03 not sure if you have any thoughts here? I'd like to allow us to start iterating on a more "public" API, but one that still should be considered experimental.

tlambert03:
Yeah I like the idea.  Will take a closer look soon and add some thoughts. 

Mostly I think it could use some docs (beyond examples)

manzt:
Agreed. I think the entire descriptor-based API needs documentation. I've been trying to find the time to write up the approach more in-depth, but minimally something in the docs (beyond examples) would be great.

manzt:
Going to merge for now. Proper documentation should be a priority before we make a release.

tlambert03:
go for it,  working on docs right now

PR #128, opened 2023-06-02
Experimental module docs [wip]

tlambert03:
docs for experimental module/descriptor

### DIFF
--- a/.changeset/pretty-rockets-behave.md
+++ b/.changeset/pretty-rockets-behave.md
@@ -1,0 +1,18 @@
+---
+"anywidget": patch
+---
+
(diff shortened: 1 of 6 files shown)


tlambert03:
ready (enough) for review @manzt  ... lemme know if you think this is a reasonable direction

manzt:
Oh wow, awesome <3 Can't wait to have a read. Also I've found the `_repr_mimbundle_` documentation is very poor elsewhere, so it's really nice to have some concrete explanation of what that is in this as well :)

manzt (approved):
Really great. Thanks for putting all this together!

PR #138, opened 2023-06-14
feat: limit view fields exposed to render function

### LINKED ISSUE #137: Prevent running view.remove() in the cleanup function or add note in docs
I did not realize that the cleanup function runs during `view.remove()`. I added a view.remove() call within the cleanup function returned by render() and then this caused an infinite loop in my widget by accident.

It might be possible to prevent users from getting in this situation in some way, but I am not sure. Maybe instead of passing `this` to render, a different object could be passed without a `remove` property or with `remove()` somehow overridden as a no-op? 

https://github.com/manzt/anywidget/blob/8e589912f7a60989b90a5683a5a35a8a67456c08/packages/anywidget/src/widget.js#L174

If preventing it is not possible, the docs about the cleanup function could note that view.remove() should not be called inside the returned function, or could clarify that the cleanup function is run during `DOMWidgetView.remove`

manzt:
## Description:

This PR addresses issue #137 by making changes to object exposed to user-provided `render`. 

Previous:

```javascript
class AnyView {
  async render() {
    // ...
    let cleanup = widget.render(this);
    // ...
  }
}
```

Proposed:

```javascript
class AnyView {
  async render() {
    // ...
    let cleanup = widget.render({ model: this.model, el: this.el });
    // ...
  }
}
```

The key focus is on minimizing the API surface area by reducing the exposed fields to the user-provided render function. This change is expected to improve encapsulation and limit the potential for erroneous modifications.

In a review of current projects utilizing anywidget, no evident use-cases were identified that would be adversely affected by this change. It's worth mentioning that there's still a degree of uncertainty about how these changes will impact model (un)subscriptions within Backbone.

## To-Do:

- [ ] Investigate the potential impacts of this change on granular model subscriptions and unsubscriptions in Backbone.
- [x] Update documentation with new types

## Request for Review:

I would like to invite reviewers to provide feedback, especially in areas related to Backbone's model (un)subscriptions. Any objections or observations on potential impacts are greatly appreciated.

cc: @maartenbreddels @juba @domoritz @kolibril13

### DIFF
--- a/packages/anywidget/src/widget.js
+++ b/packages/anywidget/src/widget.js
@@ -1,11 +1,11 @@
 // @ts-check
 import { name, version } from "../package.json";
 
-/** @typedef {import("@jupyter-widgets/base").WidgetModel} WidgetModel */
+/** @typedef {{ model: import("@jupyter-widgets/base").WidgetModel, el: HTMLElement }} RenderContext */
 
 /**
  *  @typedef AnyWidgetModule
- *  @prop render {(view: { model: WidgetModel, el: HTMLElement }) => Promise<undefined | (() => Promise<void>)>}
+ *  @prop render {(context: RenderContext) => Promise<undefined | (() => Promise<void>)>}
  */
 
 /**
--- a/packages/anywidget/src/widget.js
+++ b/packages/anywidget/src/widget.js
@@ -1,11 +1,11 @@
 // @ts-check
 import { name, version } from "../package.json";
 
-/** @typedef {import("@jupyter-widgets/base").WidgetView} WidgetView */
+/** @typedef {import("@jupyter-widgets/base").WidgetModel} WidgetModel */
 
 /**
  *  @typedef AnyWidgetModule
- *  @prop render {(view: WidgetView) => Promise<undefined | (() => Promise<void>)>}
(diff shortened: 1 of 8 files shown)


packages/anywidget/src/widget.js
```diff
@@ -1,11 +1,11 @@
 // @ts-check
 import { name, version } from "../package.json";
 
-/** @typedef {import("@jupyter-widgets/base").WidgetView} WidgetView */
+/** @typedef {import("@jupyter-widgets/base").WidgetModel} WidgetModel */
 
 /**
  *  @typedef AnyWidgetModule
- *  @prop render {(view: WidgetView) => Promise<undefined | (() => Promise<void>)>}
+ *  @prop render {(view: { model: WidgetModel, el: HTMLElement }) => Promise<undefined | (() => Promise<void>)>}
```
domoritz, line 8: Is it odd that the variable is still called `view` here?

domoritz (approved):
Works for my use case.

packages/anywidget/src/widget.js
manzt: Ah yes, that's a good suggestion. Maybe a type like `RenderContext` or `RenderProps` would be better (variable names `context` or `props` respectively)

juba:
This would work for my use case, too.

PR #140, opened 2023-06-16
feat: restrict access to model within render context

manzt:
Similar to #138

Proxies access to `AnyModel` to automatically provide the `AnyView` instance as context to `model.on` and `model.off`. This ensures that the event handlers will be properly cleaned up when the view is unmounted by anywidget.

### DIFF
--- a/packages/anywidget/src/widget.js
+++ b/packages/anywidget/src/widget.js
@@ -171,7 +171,12 @@ export default function ({ DOMWidgetModel, DOMWidgetView }) {
 					let widget = await load_esm(this.get("_esm"));
 
 					// call any cleanup logic defined by the previous module.
-					await view._anywidget_cached_cleanup();
+					try {
+						await view._anywidget_cached_cleanup();
+					} catch (e) {
+						console.warn("[anywidget] error cleaning up previous module.", e);
+						view._anywidget_cached_cleanup = () => {};
+					}
 
 					// Remove all event listeners added by the user-defined render.
 					this.off(null, null, view);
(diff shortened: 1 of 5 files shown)


packages/anywidget/src/widget.js
```diff
@@ -98,6 +98,39 @@ async function load_esm(esm) {
...
+ * @param {import("@jupyter-widgets/base").DOMWidgetView} view
+ * @returns {import("@anywidget/types").RenderContext}
+ *
+ * Prunes the view down to the minimum context necessary for rendering.
+ * Calls to `model.get` and `model.set` automatically add the view as
+ * context to the model, so we can gracefully unsubscribe from events
+ * added by the user-defined render.
+ */
+function extract_context(view) {
+	let model = {
```
nvictus, line 111: lgtm

I suppose you could "leak" the `AnyView` instance as a private `_view` member of the model proxy in case there is exceptional use case for it, but it might make more sense to wait for someone to ask for functionality the model proxy can't provide.

manzt: I like the idea of waiting until/if there is a use case to expose the view in the render context. With this PR in place, very few jupyter-widget/backbone things are exposed to end users. It's a much narrower API that we need to support, which also allows us greater flexibility to change internals. 

For example, I have an idea for how to support custom serializers with this API.

domoritz: This removes model.send which I need. Can you bring that back?

manzt: Oh no, sorry about that! Fixed in #146 and released in v0.5.2.

PR #167, opened 2023-06-27
feat: bring back support for Python 3.7

### LINKED ISSUE #166: Sync Python support with ipywidgets?
v0.6 dropped support for Python 3.7. Perhaps it was a bit [too aggressive](https://twitter.com/maartenbreddels/status/1673426929939107841?s=46&t=gbRQ9_X7aXunEwPKYKknsw)? My motivation for dropping support is because 3.7 is now officially EOL, but keeping support longer shouldn't be super complicated. Maybe a good plan would be to defer to whatever ipywidgets is supporting?

manzt:
Fixes #166

### DIFF
--- a/packages/anywidget/package.json
+++ b/packages/anywidget/package.json
@@ -34,7 +34,7 @@
 	},
 	"devDependencies": {
 		"@jupyter-widgets/base-manager": "^1.0.5",
-		"@jupyterlab/builder": "^4.0.2"
+		"@jupyterlab/builder": "^3.6.5"
 	},
 	"jupyterlab": {
 		"extension": "src/plugin",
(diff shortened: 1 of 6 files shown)


packages/anywidget/package.json
```diff
@@ -34,7 +34,7 @@
 	},
 	"devDependencies": {
 		"@jupyter-widgets/base-manager": "^1.0.5",
-		"@jupyterlab/builder": "^4.0.2"
+		"@jupyterlab/builder": "^3.6.5"
```
manzt, line 37: Really frustrating how `@jupyterlab/builder` is tied to the version of JupyterLab.

maartenbreddels:
awesome ❤️

schwehr:
bc31487d47ee6a1322d86033791c9f7b32c1a011 made python 3.8 the minimum version.

That means means the todo in widget.py's `__init_subclass__` can be resolved.

https://github.com/manzt/anywidget/blob/9b7366d7cf0b98461ab135550667c303fade7e2e/anywidget/widget.py#L65-L74

PR #174, opened 2023-07-05
feat(descriptor): support Pydantic v2

manzt:
I'm doubtful that many are using the descriptor API, so this PR deprecates support for pydantic v1. I'm curious if we should try to support v1 still.

cc: @tlambert03 not sure if you have any thoughts here


tlambert03:
sorry I missed this... but yeah, I do think you should support both.  since it's easy enough, and pydantic v1 will be around for a while

manzt:
Thanks! That was my impression.

PR #190, opened 2023-07-19
feat: add @anywidget/react package

manzt:
Follow up to discussion in https://github.com/Octoframes/anywidget-react-vite-test/pull/1#issuecomment-1642180284.

Adds the `@anywidget/react` package that takes care of some boilerplate to connect Jupyter Widget's backbone models
with React.

Usage (without a bundler):

```python
import anywidget
import traitlets

    _esm = """
    import * as React from "https://esm.sh/react@18";
    import { createRender, useModelState } from "https://esm.sh/@anywidget/react?deps=react@18,react-dom@18";
    function Counter() {
      let [value, setValue] = useModelState("value");
      return React.createElement(
        "button",
        { onClick: () => setValue(value + 1) },
        `count is ${value}`
      );
    }
    export let render = createRender(Counter);
    """

Counter()
```

Usage (with bundler):

```sh
npm i react react-dom @anywidget/react
```

```javascript
import * as React from "react";
import { createRender, useModelState } from "@anywidget/react";

function Counter() {
  let [value, setValue] = useModelState("value");
  return <button onClick={() => setValue(value + 1)}>count is {value}</button>
}

export let render = createRender(Counter);
```

```sh
npx esbuild --bundle --format=esm --outfile=bundle.js counter.jsx
```

```python
import anywidget
import traitlets

    _esm = "bundle.js"

Counter()
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,1 +1,0 @@
-packages/anywidget/CHANGELOG.md
--- a/packages/react/README.md
+++ b/packages/react/README.md
@@ -1,0 +1,27 @@
+# @anywidget/react
+
+> React utilities for [**anywidget**](https://anywidget.dev)
+
+## Installation
+
+```sh
+npm install @anywidget/react
+```
+
+## Usage
+
+```javascript
+import * as React from "react";
+import { createRender, useModelState } from "@anywidget/react";
+
+function Counter() {
+  let [value, setValue] = useModelState("value");
+  return <button onClick={() => setValue(value + 1)}>count is {count}</button>
+}
+
+export let render = createRender(Counter);
+```
+
+## License
+
+MIT
--- a/packages/react/README.md
+++ b/packages/react/README.md
@@ -15,8 +15,8 @@ import * as React from "react";
 import { createRender, useModelState } from "@anywidget/react";
 
 function Counter() {
-  let [value, setValue] = useModelState("value");
-  return <button onClick={() => setValue(value + 1)}>count is {count}</button>
+	let [value, setValue] = useModelState("value");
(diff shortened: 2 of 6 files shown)


manzt:
Maybe we could standardize an adapter interface that would allow greater control of modules.

```javascript
import { adapter, useModelState } from "@anywidget/react";

function App {

}

export default adapter(Widget);
```

manzt:
cc: @keller-mark 

Curious what you think of this.

keller-mark:
This looks great. Only minor thing is how to provide a custom cleanup function if using `createRender`?

Another parameter for `createRender` could be added for this, but instead I would expose an underlying `ModelProvider` component. I know it is less convenient to use as it requires an extra level of component but it is more appealing to me to stay more react-ish and an option to keep the control over the ReactDOM usage.

Perhaps like
```
function InnerWidget() {
  const [value] = useModelState('key');
  // ...
}

function OuterWidget({ model }) {
  return (
  <ModelProvider model={model}>
    <InnerWidget>
  </ModelProvider>
}

function render(view) {
  // render <OuterWidget model={view.model} /> into view.el
}
```

manzt:
I'm not 100%, but couldn't you use a `useEffect` to cleanup? I think that will be called when the component is unmounted (which happens in the cleanup function).

```javascript
import * as React from "react";
import { useModelState, createRender } from "@anywidget/react";

function Widget() {
  let [value, setValue] =useModelState("value");
  React.useEffect(() => {
    return () => {
       // cleanup logic?
    }
  }, []);
  return <div>value is {value}</div>
}

export let render = createRender(Widget);
```

keller-mark:
Ah that is a nice option that I had not considered, yes that should work too

Issue #202, opened 2023-07-24
Consolidate additional front-end helpers / framework adapters


manzt:
**Goal**: Consolidate common patterns for communicating between JS and Python.

**Why**: Someone comes to **anywidget** and wants to use framework _X_. How do you make sure they can get the information they need and be productive as quickly as possible.

**How**: Perhaps some additional utility libraries (a la `@anywidget/react`) that provide the minimal boilerplate to make synchronizing state between JS & Python a breeze. @xnought has a really nice svelte implementation in https://github.com/xnought/svelte-store-anywidget, for example. I guess the question is how we should share this kind of code:

1. Official libraries (i.e.,`@anywidget/react`, `@anywidget/svelte`)
2. Snippets or recipes that can be copied and pasted
3. An awesome-anywidget collection of cool anywidget projects

Out of the above options, I think I'd prefer 1. Since many coming to anywidget for the first time might not be familiar with front-end tools / JS, having an official "one-stop-shop" for your anywidget needs would be the best experience IMO. Perhaps we could even have some templates like popular frontend frameworks to bootstrap any anywidget project:

```sh
npm create anywidget@latest my_widget --tempate=react
```

Biggest question on my mind is maintenance. I think the client-specific libraries shouldn't be too big of a lift, but the templates would be really tricky to keep up to date (**I do not want cookiecutters part 2**). 

**Musing on project philosophy**: One thing I like about **anywidget** in its current state is that the starting point is ESM. For beginners, this means you start by handwriting JS code that the browser understands and thereby get a better understanding and demystify of the front end. Front-end frameworks introduce magic (like bundlers) that obscure this fact, meaning the code you write is NOT what is executed in the browser, which can lead to misconceptions / confusion about front end. 

Ideally one would only reach for a framework if the abstractions it provided outweighed the cost (often not a consideration), but the reality is that many come to the front-end for the first time having heard about framework _X_ and now they want to use _X_ for their first project. 

IMO widgets should primarily utilize vanilla JS, with the majority of logic to connect Jupyter Widget's backbone `model` with an existing API. If a widget is sufficiently complex and necessitates a front-end framework, it likely signifies a more general front-end library that would be better off in a separate package. 

With that said, I think folks will still come to anywidget and want to use framework _X_, so I want to make sure they avoid common pitfalls and are productive quickly.

manzt:
Implemented for react/svelte now. We can add more in the future.

PR #205, opened 2023-07-25
feat: add `@anywidget/svelte` adapter

manzt:
Towards #202 

Proof of concept, inspired by https://github.com/xnought/svelte-store-anywidget.

Not sure my `onDestroy` behavior will clean things up correctly. Maybe we could keep WeakMap/WeakSet to keep track of stores without requiring users to explicitly cleanup. Thinking the Proxy stuff might be a little too much magic, but I like the restructuring API (you get autocomplete with typescript).

Usage:

```svelte
<script>
    import { stores } from "@anywidget/svelte";
    const { count } = stores;
</script>

<button on:click={() => $count += 1}Count is {$count}</button>
```

```js
import { createRender } from "@anywidget/svelte";
import Counter from "./Counter.svelte";

export const render = createRender(Counter);
```

The `Stores` also is generic over a Model object for those using TypeScript. I took this casting pattern from astro.

```svelte
<script lang="ts">
    import { stores, type Stores } from "@anywidget/svelte";
    const { count } = stores as Stores<{ count: number }>;
</script>

<button on:click={() => $count++}Count is {$count}</button>
```

Curious what you think @xnought @gereleth. Definitely a prototype and would love to find a solution that is most sveltey.

Update: rather than react-like hooks the `stores` and `model` are exposed as constants to the child module.

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -12,7 +12,7 @@
 		}
 	},
 	"scripts": {
-		"build": "tsc --declaration --emitDeclarationOnly --declarationMap --outDir dist",
+		"build": "tsc --declaration --emitDeclarationOnly --outDir dist",
 		"typecheck": "tsc --noEmit"
 	},
 	"peerDependencies": {
--- a/packages/react/tsconfig.json
+++ b/packages/react/tsconfig.json
@@ -1,8 +1,3 @@
 {
-	"extends": "../../tsconfig.json",
-	"compilerOptions": {
-		"moduleResolution": "nodenext",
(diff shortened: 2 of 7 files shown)


xnought:
This is epic. Looks simple and useful. I like it!

manzt:
Slightly updated the syntax. Now I just export a const called `stores` rather than a getter.

manzt:
Made an initial release (very experimental API)! https://www.npmjs.com/package/@anywidget/svelte

PR #213, opened 2023-08-01
feat(svelte): use Context API to share `model`/`stores` between components

### LINKED ISSUE #207: Error while running the svelte/store example
I'm giving a try to this very new feature https://github.com/manzt/anywidget/tree/main/packages/svelte but my example fails with `count` having `NAN` value.
The svelte code is exactly the same as in the readme. The python code is like below

```
class SvelteCounter(anywidget.AnyWidget):
    _esm = "http://localhost:5173/dist/index.js?anywidget"
    count = traitlets.Int().tag(sync=True)
    
counter = SvelteCounter()
counter
```

manzt:
Fixes #207 

Uses the Svelte [Context API](https://svelte.dev/tutorial/context-api).


manzt:
still doesn't seem to work with Vite plugin.

PR #246, opened 2023-09-01
Making create-anywidget bulletprooved

kolibril13:
Follow up from https://github.com/Octoframes/anywidget-minimal-react-template/issues/3

In the old version of create-anywidget, copying template files  and then renaming them in the target dir was breaking the git repo.

With this new approach "Read, save to json, modify json, write to the file system", this error does not occur any more
Tested this via:

mkdir hi && cd hi && git init && touch hello.py && echo 'print("hello")' > hello.py
git add . && git commit -m "first file"
node /Users/jan-hendrik/projects/anywidget/packages/create-anywidget/index.js hi
and that works now!

### DIFF
--- a/packages/create-anywidget/create.js
+++ b/packages/create-anywidget/create.js
@@ -20,7 +20,6 @@ export async function create(target, options) {
   const updatedContentFiles = replaceWidgetName(updatedFiles, options);
 
   const updatedPathFiles = updateFilePaths(updatedContentFiles, copyFrom, copyTo);
-  console.log(copyTo);
   try {
     await createFiles(updatedPathFiles);
     console.log("All files created successfully!");
--- a/packages/create-anywidget/create.js
+++ b/packages/create-anywidget/create.js
@@ -63,7 +63,12 @@ function replaceWidgetName(files, newName) {
 function updateFilePaths(files, sourceDir, destDir, newName) {
   return files.map((file) => {
     let newPath = file.path.replace(sourceDir, destDir);
-    newPath = newPath.replace(/my_widget/g, newName); // Replace folder name
+    
+    // Check if path has the structure 'src/my_widget'
+    if (newPath.includes('src/my_widget')) {
+      newPath = newPath.replace(/src\/my_widget/g, 'src/' + newName);
+    }
+
     file.path = newPath;
     return file;
   });
--- a/packages/create-anywidget/create.js
+++ b/packages/create-anywidget/create.js
@@ -7,20 +7,20 @@ export async function create(target, options) {
   const __dirname = path.dirname(url.fileURLToPath(import.meta.url));
   const copyFrom = path.resolve(__dirname, options.template);
   const copyTo = target;
-  const newName = snakecase(options.name); 
+  const newName = snakecase(options.name);
 
   const allFiles = await gatherFiles(copyFrom);
   const updatedFiles = renameGitignore(allFiles);
-  const updatedContentFiles = replaceWidgetName(updatedFiles, newName); // Pass newName as argument
-  const updatedPathFiles = updateFilePaths(updatedContentFiles, copyFrom, copyTo, newName); // Pass newName as argument
+  const updatedContentFiles = replaceWidgetName(updatedFiles, newName);
+  const updatedPathFiles = updateFilePaths(
+    updatedContentFiles,
+    copyFrom,
+    copyTo,
+    newName
+  );
 
-  console.log(copyTo);
-  try {
-    await createFiles(updatedPathFiles);
-    console.log("All files created successfully!");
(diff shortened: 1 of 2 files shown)


kolibril13:
@manzt : I think this is now ready to review! :)

manzt:
Thank you! I'll try to have a look over the weekend :)

packages/create-anywidget/create.js
```diff
@@ -1,70 +1,86 @@
...
-		} else {
-			throw Error("unknown type");
-		}
-	}
+  console.log(copyTo);
+  try {
+    await createFiles(updatedPathFiles);
+    console.log("All files created successfully!");
+  } catch (err) {
+    console.error("Error writing files:", err);
```
manzt, line 22: I'd like to avoid logging from these functions.

we should just let `createFiles` throw and error and then handle that error in `index.js` (e.g., `process.exit(1)`).

manzt:
```javascript
return createFiles(updatedPathFiles)
```

manzt: or actually, why not write the files and then return an array of the written paths (if successful). Then we can (optionally) log those files in `index.js`.

```javascript
await writeFiles(updatedPathFiles);
return updatedPathFiles.map(file => file.path);
```

packages/create-anywidget/create.js
```diff
@@ -1,70 +1,86 @@
...
+    return file;
+  });
+}
+
+async function createFiles(files) {
+  const writePromises = files.map(async (file) => {
+    await fs.mkdir(path.dirname(file.path), { recursive: true });
+    // Write (or overwrite) the file
+    await fs.writeFile(file.path, file.content, "utf-8");
+    console.log(`Created: ${file.path}`);
```
manzt, line 82: Again, I'd like to avoid logging from these functions.

packages/create-anywidget/create.js
```diff
@@ -1,70 +1,86 @@
...
+  });
+
+  return Promise.all(writePromises);
```
manzt, line 85: No need to return these promises

```suggestion
  await Promise.all(writePromises);
```

packages/create-anywidget/index.js
```diff
@@ -22,56 +22,56 @@ ${grey(`create-anywidget version ${pkg.version}`)}
...
-		});
-		// bail if `force` is `false` or the user cancelled with Ctrl-C
-		if (force !== true) {
-			process.exit(1);
-		}
-	}
+  if (fs.readdirSync(cwd).length > 0) {
+    let force = await p.confirm({
+      message:
+        "Directory not empty. Continue? No files will be overwritten, and the git archive won't be touched.",
```
manzt, line 39: Is this true? If I had `foo` and a already `src/foo/__init__.py` in my file system, wouldn't we blindly override?

packages/create-anywidget/index.js
```diff
@@ -81,20 +81,20 @@ let i = 1;
...
+  console.log(`  ${i++}: ${bold(cyan(`cd ${relative}`))}`);
 }
 
 console.log(`  ${i++}: ${bold(cyan("npm install"))} (or pnpm install, etc)`);
 
 // prettier-ignore
 console.log(
 	`  ${i++}: ${
-		bold(cyan('git init && git add -A && git commit -m "Initial commit"'))
+		bold(cyan('git add -A && git commit -m "add widget infrastructure"'))
```
manzt, line 92: I'd prefer the original message. This makes sense if you're adding to an existing repo, but the previous is for bootstrapping a new repo. I guess we could decide between the two, but for now I'd just prefer to give the instructions for creating a new git repot and commiting.

```suggestion
		bold(cyan('git init && git add -A && git commit -m "Initial commit"'))
```

manzt (changes requested):
Looks great! Made some small suggestions.

packages/create-anywidget/index.js
kolibril13: you're right, it will be overwritten!
I've removed that statement again.

kolibril13:
OK, I think all changes are now incorporated and this cookiecutter is now ready to use.

@manzt in a prev. conversation you said:
>  Again, want to avoid cookiecutters so not sure if this is a sustainable approach. But maybe the config is simple enough (famous last words).

What do you think right now about that?

manzt:
> What do you think right now about that?

Let's just go for it. I think for the time being it is simple enough, and a nice to have. The one thing that I'm worried about is making sure the dependencies in the templates stay up to date (e.g., react/react-dom/@anywidget/react).

Moving forward, I'd like to have a "vanilla" (no dependencies) option and svelte.

manzt (approved):

PR #256, opened 2023-09-03
move package json to root folder

kolibril13:
I think that having package.json in the root folder is more convenient then having nested in the js folder.

When I was developing anywidget projects, I had to type `cd js` and `cd ..` a lot of times.
IMO having package json directly in root would make developing and installing packages easier.


manzt (approved):
Looks great! Would you mind adding a changeset (pnpm changeset, and follow the prompts).

kolibril13:
nice to hear! 
sure, I did `npx changeset` and got:

image

was that the right way?

I also have some further template tweaks (adding an example notebook and making "build": "esbuild --minify --format=esm --bundle --outdir=src/my_widget widget.js" ) more general by adding the wildcard *js , but I think it would be better to make that later today in a sperate pr.

packages/create-anywidget/package-lock.json
manzt, line 1: oops, I think this was checked in by accident.

.changeset/chilly-zebras-appear.md
```diff
@@ -0,0 +1,5 @@
+---
+"create-anywidget": major
```
manzt, line 2: We automate releases (and versioning) with [changesets](https://github.com/changesets/changesets). When we create a release, changesets collects all the .md files from `.changeset` and determines which packages need to be released and with what versions.

For this kind of CLI, I don't think versioning is super important (what is a "breaking" change) but I'd probably make changing the location of `js` a minor changeset (since we are pre-1.0).

```suggestion
"create-anywidget": minor
```

.changeset/chilly-zebras-appear.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+move package.json to root folder
```
manzt, line 5: Also, the text that goes in this file will end up in the final CHANGELOG for `create-anywidget` when we make a release ([Example](https://github.com/manzt/anywidget/releases/tag/create-anywidget%400.1.2)), so I typically like to have these messages a bit more verbose.

```suggestion
feat: move `package.json` to template root directories

This makes it more ergonomic to install JavaScript dependencies and start the development server.
```

manzt:
> was that the right way?

Yup, but we probably don't need this to be the first major release. You can select major/minor/patch in the CLI before writing the changeset.

.changeset/chilly-zebras-appear.md
kolibril13: awesome, thanks for your advices on package management!

packages/create-anywidget/package-lock.json
kolibril13: Yep, I've removed it!

kolibril13:
Thanks for the review, I've incorporated all the changes now!

manzt:
> I also have some further template tweaks 

Great! Before sending PRs, maybe we could discuss. The larger the templates become the more unwieldy it will be to make sure all of them stay in sync. Right now this is very manual and maybe we can think of some ways to unify/simplify the templates. For example, the `pyproject.toml` and Python code are the same regardless of the template (basically only `js/` dir changes and `package.json`). 

Maybe we could have a standard layout for all the Python-stuff (e.g., so if we tweak the `pyproject.toml` all of the other templates get the same changes automatically).

> adding an example notebook

I'm open to this idea, but I'd like the templates to be very minimal. The thing I really disliked about the cookiecutter templates is that they are so "batteries included" and there are so many files that It's hard to know what's going on (e.g., lots of linting/formatting/testing stuff for the JS). 

FWIW, you don't need a Jupyter Notebook to develop a widget, and if you wanted one it would just have:

```python
from my_widget import Counter()
Counter()
```

In VSCode, you can just run `Jupyter: Create Interactive Window` and start working:

https://github.com/manzt/anywidget/assets/24403730/a352aeb4-b734-40df-a0f6-413d221bc7fa

> making "build": "esbuild --minify --format=esm --bundle --outdir=src/my_widget widget.js" ) more general by adding the wildcard *js

I like the current explicitness in this case. I agree that *js is more general, but you probably want that if you have multiple widgets in the same repo (which the rest of the template isn't designed for). Plus, esbuild supports many different file extensions that are in the wild (i.e., `.mjs`, `.cjs`, `.jsx`, `.tsx`, `.ts`, `.mts`, `.cts`) and this general solution might then fail unexpectedly.

kolibril13:
> Great! Before sending PRs, maybe we could discuss

I totally agree with you, also like things to be as simple as possible!
Do we want to have a voice call some time? I've sent you a DM on twitter!

kolibril13:
regarding an example notebook in the cookie cutter ->
> I'm open to this idea, but I'd like the templates to be very minimal. The thing I really disliked about the cookiecutter templates is that they are so "batteries included" and there are so many files that It's hard to know what's going on (e.g., lots of linting/formatting/testing stuff for the JS).

> FWIW, you don't need a Jupyter Notebook to develop a widget, and if you wanted one it would just have:
>  ```
> from my_widget import Counter()
> Counter()
> ```

I think it would be good to have that notebook called `example.ipynb` in the root folder with this content
```py
from my_widget import Counter()
Counter()
```
Even though it's just two lines, it helps to bring structure to the project, and we can also customize the package name from `my_widget` to what the name of the project actually is. 
We can also assume that people come from the JavaScript ecosystem and might have never used a notebook before.
In that case, it might be non-trivial for them to create a new notebook.

manzt:
Happy to accept a PR.

PR #260, opened 2023-09-03
fix(create-anywidget): esbuild entry paths

manzt:


kolibril13 (approved):
Thanks for fixing the path!
I also just noted that I forgot to update this here.

Issue #255, opened 2023-09-03
Cannot install package.json from template


kolibril13:
I get the following error:
jan-hendrik@JanHendriksAir okapi4 % cd js
jan-hendrik@JanHendriksAir js % npm install
npm ERR! code EUNSUPPORTEDPROTOCOL
npm ERR! Unsupported URL Type "workspace:": workspace:^

package.json
```
"dependencies": {
"@anywidget/react": "workspace^",
"react": "^18.2.0",
"react-dom": "^18.2.0"
},
```

PR #263, opened 2023-09-03
feat(create-anywidget): simplify project templates

manzt:
- feat: new template engine
- single-template engine
- consolidate scripts
- cleanup
- changeset


kolibril13:
Heeyy @manzt thanks for these changes!
I was actually also thinking of making it this way too but thought it would take too much time to implement.
Cool that you've made this in only a few hours! 


just a minor bug that I ran into while testing 
( same as https://github.com/manzt/anywidget/issues/255 )
```
jan-hendrik@JanHendriksAir okapi4 % npm i 
npm ERR! code EUNSUPPORTEDPROTOCOL
npm ERR! Unsupported URL Type "workspace:": workspace:^
```

manzt:
This error only happens if you don't have a "production" version of create-anywidget locally. If you run npm create anywidget@latest the deps are correct (no workspaces). 

The versions for the template dependencies are grabbed from the create-anywidget/package.json devDependencies. When we publish, the correct versions are updated in the package.json via pnpm. This is useful because we can rely on pnpm and GitHub dependabot to help us keep template deps up to date (no manual steps).

To prototype, you can change the version in devDependencies from "workspace" to whatever the latest version is on npm. Alternatively, you can run 'pnpm pack' in create-anywidget and un-tar the tarball for npm.

manzt:
We could maybe shell out locally to find packages for the templates:

```typescript
/** @param {{ dependencies: string[], dev_dependencies: string[]}} template */
async function get_dependency_versions(template) {
	let rook_pkg = await fs
		.readFile(path.join(__dirname, "package.json"), "utf-8")
		.then(JSON.parse);

	let lookup = rook_pkg.devDependencies;
	if (Object.values(lookup).some((v) => v.startsWith("workspace:"))) {
		lookup = { ...lookup, ...workspace_overrides() };
	}

	/**
	 * pnpm will help us keep package versions in sync over time, along with dependabot.
	 * so we lookup the version from there for any dependencies in our templates.
	 * @param {string[]} names
	 * @returns {Record<string, string>}
	 */
	function gather_dependencies(names) {
		/** @type {Record<string, string>} */
		let deps = {};
		for (let name of names) {
			let version = lookup[name];
			if (!version) {
				throw new Error(
					`No version found for ${name}. Must add to create-anywidget/package.json.`,
				);
			}
			deps[name] = version;
		}
		return deps;
	}

	return {
		dependencies: gather_dependencies(template.dependencies),
		devDependencies: gather_dependencies(template.dev_dependencies),
	}
}

/** @returns {Record<string, string>} */
function workspace_overrides() {
	let out = child_process.execSync("pnpm m ls --json --depth=-1", {
		encoding: "utf-8",
	});
	/** @type {{ name: string, version: string }[]} */
	let workspace_pkgs = JSON.parse(out);
	return Object.fromEntries(
		workspace_pkgs.map((pkg) => [pkg.name, pkg.version]),
	);
}

```

manzt:
Implemented in #265

PR #269, opened 2023-09-11
feat(create-anywidget): Support Bun

manzt:
Bun is a replacement for Node.js and comes with a built-in package manager and bundler (and test runner!). I think it very well suited as a companion for anywidget projects. I'd like the templates to fully support bun.

This PR detects if the project is running in Node.js or bun by looking at the `globalThis.Bun` and then generates the `build` script in the package.json based on the the environtment:

- Node.js - `esbuild --bundle`
- Bun - `bun bundle`

The difference is that this eliminates a dependency in those templates in favor of the "bun" version.

TODO: I need to figure out how/if the hatchling jupyter builder supports bun.

### DIFF
--- a/.changeset/small-spies-sort.md
+++ b/.changeset/small-spies-sort.md
@@ -1,0 +1,11 @@
+---
+"create-anywidget": minor
+---
+
+feat: Support Bun and prefer built-in bundler over `esbuild`
+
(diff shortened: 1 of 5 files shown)


manzt:
Seems to be a blocker at the moment: https://github.com/oven-sh/bun/issues/4835

PR #273, opened 2023-09-15
Add .venv to gitignore

kolibril13:
For convenience.

I always include .venv in my global gitignore. However, I discovered that "hatch build" requires .venv to also be in the project's gitignore. If not, .venv will be uploaded to PyPI. I experienced this issue firsthand; it took a long time to upload, and it was a challenge to identify the reason for the large package size!


kolibril13 (commented):
Another tweak:
only-packages = true
only puts the package on pypi, not other example folders.

manzt:
Looks great! can you update the snapshot for our tests? `pnpm test` from the root.

PR #274, opened 2023-09-15
feat: override DOMWidgetModel.serializer to support binary data

manzt:
The recent release of ipywidgets breaks existing anywidgets with binary traits. This PR overrides the serializer to enforce the use of `structuredClone` and binary data.


manzt:
@hamelin, this was part of the selection bug you were experiencing at SciPy and we will fix it here for jupyter-scatter.

hamelin:
> @hamelin, this was part of the selection bug you were experiencing at SciPy and we will fix it here for jupyter-scatter.

Great job!

PR #311, opened 2023-10-01
accept imports in deno anywidget

rgbkrk:
This is a rough way to include imports, which could be a general preamble instead.

```typescript
import { widget } from "./mod.ts";

let model = await widget({
    // shared model between frontend and backend
    state: { data: [4, 8, 15, 16, 23, 42] },
    imports: `
import * as d3 from "https://esm.sh/d3@7";
`,
    // front-end render function 
    render: ({ model, el }) => {
        const width = 500;
        const svg = d3.select(el)
          .append("svg")
          .attr("width", width)

        const x = d3.scaleLinear()
          .range([0, width]);

        const updateChart = (data) => {
            // Update scale domain based on new data
            x.domain([0, d3.max(data)!]);

            const bars = svg.selectAll("rect")
                .data(data);

            // Update existing bars
            bars
                .attr("width", x)
                .attr("y", (d, i) => i * 25)
                .attr("height", 20);

            // Add new bars
            bars.enter().append("rect")
                .attr("x", 0)
                .attr("y", (d, i) => i * 25)
                .attr("width", x)
                .attr("height", 20)
                .attr("fill", "#69b3a2");

            // Remove extra bars
            bars.exit().remove();
        };

        // Initial render
        updateChart(model.get('data'));

        model.on("change:data", () => {
            updateChart(model.get('data'));
        });
    }
});

model
```

Export-1696124360731

On a side note, my editor defaults to using `deno fmt`. Could we use the builtin deno formatting for the deno project portion?

### DIFF
--- a/packages/anywidget/mod.ts
+++ b/packages/anywidget/mod.ts
@@ -90,8 +90,9 @@ type FrontEndModel<State> = Model<State> & {
 	save_changes(): void;
 };
 
-export async function widget<State>({ state, render }: {
+export async function widget<State>({ state, render, imports }: {
 	state: State;
+	imports: string;
 	render: (
 		context: {
 			model: FrontEndModel<State>;
@@ -103,9 +104,10 @@ export async function widget<State>({ state, render }: {
 	let comm = new Comm();
 	await comm.init();
 	// TODO: more robust serialization of render function (with context?)
+	const _esm = `${imports}\nexport const render = ${render.toString()}`;
 	await comm.send_state({
 		...state,
-		_esm: "export const render = " + render.toString(),
+		_esm,
 	});
 	for (let key in state) {
 		model.on(`change:${key}`, (data) => {


rgbkrk:
Made it work with React and https://github.com/denoland/deno/pull/20695. Since `createRender` relies on variables being defined where the `render` is run, it has to be called "on the frontend" rather than be created in Deno.

image

rgbkrk:
This is too great.

confetti

rgbkrk:
I'm liking this interface for React based anywidgets 

image

manzt (approved):
Awesome! Thanks for the contribution. Let's merge and we can iron out react usage.

> On a side note, my editor defaults to using `deno fmt`. Could we use the builtin deno formatting for the deno project portion?

I format everyting with `dprint` (which is what `deno fmt` uses under the hood) with the deno plugin. It's just I have tabs enabled. In general, I'm realizing it might be tricky to manage a dual node/deno repo.

rgbkrk:
Must be a configuration issue on my end. I made sure to keep it consistent with the repo so the change set is small. I'll dig into editor settings.

manzt:
> Must be a configuration issue on my end. I made sure to keep it consistent with the repo so the change set is small. I'll dig into editor settings.

Thank you! I'll try my best to get it so that `deno fmt` just works too.

manzt:
Do you have any thoughts on how to publish to denoland the monorepo? We make separate tagged releases for each sub-package. I went through the setup on https://deno.com/add_module, but I think I may have made a mistake with the subpath and not sure how to modify.

rgbkrk:
Not sure. I haven't tried monorepos with Deno and I don't have any ideas (yet). Might be worth a post to the Deno discord: https://discord.com/servers/deno-684898665143206084

manzt:
Oh great call. Thank you!

PR #317, opened 2023-10-02
feat(experimental): Support getting partial state

manzt:
This PR introduces an required `include` parameter to `_get_anywidget_state` for selective field serialization. This can help optimize performance by avoiding serializing fields that aren't emitted in events to the front end.

This is a BREAKING change because it requires implementors of `_get_anywidget_state` to
account for `include` in the function signature.

```python
from dataclasses import dataclass, asdict
from io import BytesIO

import polars as pl
import psygnal

@psygnal.evented
@dataclass
class Foo:
  value: int
  df: pl.DataFrame

  def _get_anywidget_state(self, include: set[str] | None):
    data = asdict(self)
    if include and "df" in include:
      with BytesIO() as f:
        self.df.write_ipc(f)
        data["df"] = f.getvalue()
    else:
      del data["df"] # don't serialize df to bytes
    return data
```

Thoughts, @kylebarron?

### DIFF
--- a/.changeset/serious-panthers-battle.md
+++ b/.changeset/serious-panthers-battle.md
@@ -2,4 +2,32 @@
 "anywidget": patch
 ---
 
-feat(experimental): Allow picking fields when serializing state
+feat(experimental)!: Require `include` in `_get_anywidget_state` signature
+
+Allows implementors to avoid re-serializing fields which aren't needed to send to the front end.
+This is a BREAKING change because it requires implementors of `_get_anywidget_state` to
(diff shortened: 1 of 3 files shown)


manzt:
Hmm doesn't seem to be working like I thought...

manzt:
We could try to inspect the signature of the provided `_get_anywidget_state` to know whether the implementer has included `include` in the signature, but that adds a lot of complexity to an experimental API. It is more simple/easier to just require `include`, and then later we could think about relaxing it.

PR #325, opened 2023-10-05
test(deno): add `test.ts`

manzt:
@rgbkrk also exposes an `_internals` export which you can get access to the `Comm` for the `model` returned from `widget`:

```typescript
import { widget, _internals } from "./mod.ts";

let model = await widget({
  state: { value: 0 },
  render: () => {},
});

let comm = _internals.get_comm(model);
```

### DIFF
--- a/packages/deno/mod.ts
+++ b/packages/deno/mod.ts
@@ -32,7 +32,6 @@ export const _internals = {
 	},
 };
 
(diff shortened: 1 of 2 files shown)


rgbkrk:
Oooh interesting thank you.

PR #328, opened 2023-10-05
Fix Deno example README.md

rgbkrk:
The string split happens later in the example and the promise has to be awaited first.


manzt (approved):
Thanks!!

PR #329, opened 2023-10-05
feat(deno): Add (un)install scripts for front end code

manzt:
Removes the need to `pip install anywidget` for deno kernels. Can just do it with Deno:

```sh
deno run --allow-net --allow-write ./install.ts
```

TODO: Hmm also need to install `@jupyter-widgets/base`...


manzt:
cc @rgbkrk

rgbkrk:
In case it helps, there is an _old_ package on npm for jupyter-paths. https://www.npmjs.com/package/jupyter-paths

I'm tempted to just port that to pure Deno honestly...

PR #341, opened 2023-10-11
chore(deno): add unstable types to deno.json

manzt:


packages/deno/deno.json
```diff
@@ -7,5 +7,8 @@
 		"rules": {
 			"exclude": ["prefer-const"]
 		}
+	},
+	"compilerOptions": {
+		"lib": ["deno.window", "deno.unstable"]
```
rgbkrk, line 12: TIL

PR #345, opened 2023-10-13
feat(anywidget): Raise Python error when file is missing

### LINKED ISSUE #344: Confusing error when Python package is missing JS files
I moved deployment of my anywidget-based library from local to CI and forgot to build and include the relevant JS files. I got a confusing error mentioning regexes:

image

Obviously the issue here is on my side, but if it's possible to have a more descriptive error here, that would be great!

manzt:
Fixes #344. This is kind of more tricky than I'd hoped and has to deal with the overloads for `_esm`:

```python
    _esm = "http://localhost:8080/index.js" # a URL
```

```python
    _esm = "index.js" # a string for a path
```

```python
    _esm = """
    export function render({ model, el }) { /* ... */ }
    """ # string for the _contents_ of a file
```

```python
    _esm = pathlib.Path("index.js") # a pathlib.Path object
```

If we are provided a `pathlib.Path` object, we now assert that the file exists. Otherwise, we will try to "guess"
whether the string is 1.) a URL, 2.) raw file contents, or 3.) a file path.

TL;DR – an explicit `pathlib.Path` will always lead to a Python exception if the file is missing, a string for a path will _probably_ raise an exception if the file is missing.

### DIFF
--- a/.changeset/shiny-planes-burn.md
+++ b/.changeset/shiny-planes-burn.md
@@ -1,0 +1,5 @@
+---
+"anywidget": patch
+---
+
+feat: Rraise Python error when file is missing
--- a/.changeset/shiny-planes-burn.md
+++ b/.changeset/shiny-planes-burn.md
@@ -2,4 +2,4 @@
 "anywidget": patch
 ---
 
-feat: Rraise Python error when file is missing
+feat: Raise Python error when file is missing
--- a/tests/test_widget.py
+++ b/tests/test_widget.py
@@ -203,6 +203,19 @@ def test_missing_string_path_with_suffix_raises(tmp_path: pathlib.Path):
             _esm = str_path_with_suffix
 
 
+def test_remote_contents():
+    esm = "http://example.com/foo.js"
+    css = "https://example.com/bar.css"
+
+    class Widget(anywidget.AnyWidget):
+        _esm = esm
+        _css = css
+
+    widget = Widget()
+    assert widget._esm == esm
+    assert widget._css == css
+
+
 def test_missing_string_path_without_suffix_is_raw_string(tmp_path: pathlib.Path):
     str_path_without_suffix = str(tmp_path / "foo")
 
--- a/tests/test_widget.py
+++ b/tests/test_widget.py
@@ -185,23 +185,31 @@ def test_infer_file_contents(tmp_path: pathlib.Path):
     Widget._esm.stop_thread()
 
 
-def test_missing_file_no_infer(tmp_path: pathlib.Path):
+def test_missing_pathlib_path_raises(tmp_path: pathlib.Path):
     esm = tmp_path / "foo.js"
-    css = str(tmp_path / "styles.css")
+
+    with pytest.raises(FileNotFoundError):
+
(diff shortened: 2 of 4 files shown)


tests/test_widget.py
```diff
@@ -185,23 +185,44 @@ def mock_file_events():
...
-    assert Widget._css == css
+    widget = Widget()
+    assert widget._esm == esm
+    assert widget._css == css
 
-    w = Widget()
 
-    assert w._esm == str(tmp_path / "foo.js")
-    assert w._css == css
+def test_missing_string_path_without_suffix_is_raw_string(tmp_path: pathlib.Path):
```
manzt, line 219: Notably, we don't raise `FileNotFoundError` if `_esm` or `_css` are a string without a file extension. I thought `pathlib.Path` would throw for some `str` and we could use that behavior, but surprisingly:

```python
Path("""
export function foo() { }
""")
```

is totally fine, as well as:

```python
Path("""
.foo {
  background-color: red;
}
""").suffix
# .foo { ...
```

.changeset/shiny-planes-burn.md
```diff
@@ -0,0 +1,5 @@
+---
+"anywidget": patch
+---
+
+feat: Rraise Python error when file is missing
```
rgbkrk, line 5:
```suggestion
feat: Raise Python error when file is missing
```

rgbkrk (approved):
I'm personally not a big fan of letting a string be detected as either a script or a URL path. I'd rather this be a script only unless there's an equivalent URL based Pathlib Path object.

manzt:
> I'm personally not a big fan of letting a string be detected as either a script or a URL path. I'd rather this be a script only unless there's an equivalent URL based Pathlib Path object.

as in?

```python
    _esm = "http://localhost:8080/index.js" # a URL
```

I agree that the ambiguity of the string isn't desirable, but we suggest this approach for developing with Vite in [the documentation](https://anywidget.dev/en/bundling/#vite). This PR is just intended to improve the error messages in the case that a file path is provided and missing. Perhaps we could require a `pathlib.Path` in the future, but for now this keeps the current API (with hopefully a more informative error message).

kylebarron:
> I'm personally not a big fan of letting a string be detected as either a script or a URL path. I'd rather this be a script only unless there's an equivalent URL based Pathlib Path object.

This is a bit opinionated but totally agreed IMO. My own inclination would be to disallow string for either path or url input. Deducing what a string means is too magical

> as in?

My reading would be to make your own
```py
class URL(str):
   pass
```
and require a url to be passed in as an instance of `anywidget.URL`.

rgbkrk:
@manzt sorry for making a discussion out of this. I'm delighted to use this as is. Especially since this functionality is already in anywidget!

Maybe I'm really silly on this since there won't be bare strings at the start of a script that have a URL scheme. On the opposite side, I like interfaces where the string is meant to be either a local path or a URL, as is done with ESM imports and `pd.read_csv`

```python
df = pd.read_csv("states.csv")

df = pd.read_csv("https://example.org/states.csv")
```

I think I've just been doing too much strict interface code and wanted to make sure types lined up well.

PR #361, opened 2023-10-24
chore: Replace black with ruff format

manzt:


pyproject.toml
```diff
@@ -21,7 +21,6 @@ dependencies = [
 
 [project.optional-dependencies]
 test = [
-    "black[jupyter]",
```
rgbkrk, line 24: I'm new to ruff, but it looks like we can bring in [formatting for notebooks](https://docs.astral.sh/ruff/faq/#does-ruff-support-jupyter-notebooks) with:

```
[tool.ruff]
extend-include = ["*.ipynb"]
```

Issue #355, opened 2023-10-21
Version issue on anywidget for Deno is back


rgbkrk:
![image](https://github.com/manzt/anywidget/assets/836375/afa66a43-817b-4e90-98ee-660cc04837e9)

```
Error: Module anywidget, version 0.0.1 is not registered, however,         0.7.0 is
```

manzt:
Ahh! I'll try to have a look today. Thanks for reporting. This is surprising.

manzt:
@rgbkrk do you know if this is happening after pip-install anywidget or using the install script?

manzt:
Can you make sure that you are using a later version of `mod.ts` in your script?

```js
import { widget } from "https://deno.land/x/anywidget@0.0.4/mod.ts"
```

manzt:
Is the `Symbol.for("Jupyter.display")` no longer evalutated automatically for an object in Jupyter? This is kind of unfortunate.

rgbkrk:
> Is the `Symbol.for("Jupyter.display")` no longer evalutated automatically for an object in Jupyter? This is kind of unfortunate.

It should be. I'll dig in.

rgbkrk:
It appears to be working well now. Note: I'm on commit `53248e9bb3123a1b684f3f9f744bb671dfa53bc1` (JSX & TSX support on `main`).

<img width="609" alt="image" src="https://github.com/manzt/anywidget/assets/836375/622b3b48-9f46-49ae-a818-a8ae9ca2ecc0">

PR #367, opened 2023-11-01
allow Deno to check if display `in` widget

rgbkrk:
Deno does a `Symbol.for("Jupyter.display") in obj` check now instead of directly trying to get the property. This PR adds a `has` to the model proxy so that Deno will be able to display it directly (either as a result or in `Deno.jupyter.display`).


manzt:
Looks great. Still haven't been able to sort out #355.

rgbkrk:
I've been using `widget({ version: "0.7.0" })` for now.

PR #371, opened 2023-11-06
chore: Remove stability warning from README

manzt:
Anywidget is not 1.0, but this warning seems a bit agressive in its current state. Maybe we pick a different warning/phrasing before merging, but the pre-1.0 semver should be a good indicator of stability.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,9 +19,6 @@ Learn more in the
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


rgbkrk (approved):

rgbkrk:
Awesome! Thanks for creating this.

Issue #388, opened 2023-11-29
Support non-DOM widgets?


benbovy:
It seems that Anywidget currently only supports DOM widgets by wrapping ipywidgets' `DOMWidget` / `DOMWidgetModel`.  Could it also expose wrappers for `Widget` / `WidgetModel` as well?

DOM widgets likely cover 95% of the use cases but there are a few interesting applications that don't really require DOM elements like interacting with the Web audio API ([ipytone](https://github.com/benbovy/ipytone)) or the Web MIDI API.

Maybe we could just create empty DOM elements for those applications? This feels a bit hacky, though.

There is still the possibility of using ipywidgets directly but I really enjoy the anywidget layer for developing custom widgets :). Not sure if/how HMR would work and it is likely that non-browser front-ends won't work for the applications mentioned above, but I guess that using anywidgets would still make things easier for prototyping and packaging.

benbovy:
In the same vein than Anywidget's DOM widget, from the user point of view I could imagine a non-DOM anywidget look like this:

```python
import anywidget
import traitlets

# below I use the name `anywidget.AnyPureWidget` for non-DOM widgets
# would have been great to use `anywidget.AnyWidget` vs. `anywidget.AnyDOMWidget`!

class CounterWidget(anywidget.AnyPureWidget):
    _esm = """
    export default function (model) {
      let getCount = () => model.get("count");
      model.on("change:count", () => {
        console.log("count is ${getCount()}");
      });
    }
    """
    count = traitlets.Int(0).tag(sync=True)


counter = CounterWidget()
counter.count = 42    # shows message in JS console
counter               # shows Python repr (plain text)
```

Where the default exported function in the user-defined module would be called in the widget's `initialize()` method and also likely within an `on("change:_esm")` callback like for the DOM widget. I'm not familiar with anywidget internals (nor the javascript ecosystem in general) so I might be missing something, though.

If DOM widgets are used for 99% of the uses cases and non-DOM widgets for the remaining 1% that may simply be not worth having the latter available in Anywidget after all :-). Feel free to close this issue if you think this is not relevant.

manzt:
Thanks for sharing the use case! I see the usage but need to look into how Widgets without views end up getting displayed with a plain-text repr in the front end.

As a workaround for now, you could do something like display the widget plain text in a pre tag:

```python
import anywidget
import traitlets

# BaseClass
class Widget(anywidget.AnyWidget):
    
    def __init__(self, *args, **kwargs):
[...]

Issue #266, opened 2023-09-04
Lifecycle methods or non-view methods


maartenbreddels:
Hi,

right now, only a render method can be defined, which will be triggered when a view is created. It may also be useful to perform operations when a widget gets created (not a view), and when a widget gets destroyed. This allows for widgets similar to the one powering jslink etc.
Original questions that triggered this:
https://github.com/widgetti/ipyreact/issues/44
cc @giswqs

giswqs:
@maartenbreddels Thank you for looking into this. There are so many JavaScript plugins available for leaflet. It would be a game-changer if we can make those plugins work with ipyleaflet through anywidget/ipyreact.

manzt:
Thanks for opening the issue. This is something that I've been thinking about for some time and a use case I want to support, but I haven't settled on an API because I want to make sure we can handle other use cases. Thanks for sharing the one from @giswqs!

One thing that I'd been considering is something like `setup` (name TBD):

```js
export function setup() {
   // runs when model is created
   return {
     data: "...",
     foo: 100,
   };
}

export function render({ model, el }, context) {
  let { data, foo } = context;
}
```

A pattern I've observed in Jupyter Widgets is that there is some "data" that often gets appended to the `DOMWidgetModel` on initialization, and then views expect/mutate this data (but not with the on/off methods). So I've been considering an API where a `setup` function can do some initialization logic, but then also create some data that can be referenced mutated by other views.

I'm a bit nervous about passing around data this way and introducing a confusing mental model. So, perhaps to start we can think about just introducing a "setup" hook and later decide if it can return any state to pass around to views.

PR #395, opened 2023-12-02
feat(anywidget): Introduce front-end widget lifecycle methods

manzt:
EDIT: 2024-01-26

> [!WARNING]  
> This PR adds a deprecation notice for existing named `render` exports (i.e., all existing widget implementations). I'm not sure we will deprecate in the future, but migrating away from this would allow us more flexibility in loading widget ESM in the future. Migration:
>
> Remove
> ```
> export function render({ model, el }) { ... }
> ^^^^^^
> ```
>  Use
> ```
> function render({ model, el }) { ... }
>          ^^^^^^
> export default { render }
>                  ^^^^^^
> ```

**TL;DR**- The preferred way to define a widget front-end code is with a `default` object export. This object may include `render` function or **new** `initialize` function. 

```js
export default {
  initialize({ model }) {
    model.on("change:value", () => console.log(model.get("value")));
  },
  render({ model, el }) {
    model.on("change:value", () => el.innerText = model.get("value"))
  }
}
```

Combined, these methods introduce lifecycle "hooks" for widget developers when implementing front-end code.

- `initialize`: is executed _once_ when the widget front end first loads. The function has access to the `model` (no `el`), which can be set to setup event handlers / state to _share_ across views created in `render`.
- `render` is executed per _view_, or each time an output cell is rendered in a notebook. This function has access to the `model` _and_ a unique `el` for the output area. This method should be familiar, and can setup event handlers / access state specific to that view.

The default export can be an object, or a function that returns an object. This can be useful for setting up some initial state for the widget:

```javascript
export default function () {
  const localState = {};
  return {
    instantiate({ model }) { /* ... */ },
    render({ model, el }) { /* ... */ },
}
```

Towards:
 - #266
 - #388 
 
## Motivation

In the MVC framework used by Jupyter Widgets, right now anywidget treats Python as the sole "source of truth" for defining the model and only supports rendering _views_ of that model in JS (view `render`). We don't currently have an API to initialize some initial front-end model listeners/state (or allow for a DOM-less use of anywidget).

Surveying existing Jupyter Widgets in the wild, it seems this type of model initialization is often defined in the `WidgetModel.initialize` method.  This PR introduces a new lifecycle method which maybe be implemented for an anywidget front-end module: `initialize`. Initialize has the same signature as render, except only `model` is provided in the context.

The semantic difference is that `initialize` is executed _once_ per model (regardless of whether anything is displayed) and `render` is executed once per view. 

### Example 1: Model-level "logger" for state changes

Here, `initialize` adds an event listener to `console.log` any time value changes, and then each view adds their own event listener for updating the text node in the DOM. If the `console.log` were to be setup per-view, then the console would print "Value is X" for every view. Adding the handler in `initialize` ensures this logic is only handled once per instance.

```python
import anywidget
import traitlets

    _esm = """
   async function initialize({ model }) {
        model.on("change:value", () => {
            console.log(`Value is ${model.get("value")}!`);
        });
    }
    function render({ model, el }) {
        el.innerText = `Value is ${model.get("value")}!`;
        model.on("change:value", () => {
            el.innerText = `Value is ${model.get("value")}!`;
        })
    }   
    export default { initialize, render }
    """

widget = Widget()
widget
```

```python
widget
```

```python
widget.value = 10 # logs once to the browser console and updates both DOM views
```

### Example 2: DOM-less widget

```python
import anywidget
import traitlets

    _view_name = traitlets.Any(None).tag(sync=True) # TODO: nicer way to do this?
    _esm = """
   async function initialize({ model }) {
        model.on("change:value", () => {
            console.log(`Value is ${model.get("value")}!`);
        });
    }
    export default { initialize }
    """

widget = Widget()
widget.value = 1 # logs in the browser console
widget # displays a plain/text repr
# Widget()
```


manzt:
I'm thinking I will likely get rid of keeping the `model._unstable_context`. The same effect can be accomplished within the module itself with a separate variable:

```javascript
const value_history = [];

export async function _unstable_setup({ model }) {
	model.on("change:value", () => {
		value_history.push(model.get("value"));
	});
}
export function render({ model, el }) {
	let button = document.createElement("button");
	button.innerHTML = `Log history`;
	button.addEventListener("click", () => {
		console.log(`the count history is ${value_history}`);
	});
	el.appendChild(button);
}
```

manzt:
I have renamed this method from `setup` to `initialize` to mirror the method it represents in traditional Jupyter Widgets (`WidgetModel.initialize`), much like how `render` reflects `DOMWidgetModel.render`.

maartenbreddels:
On my first review this was the only comment I had :)

> `setup` to `initialize` to mirror the method it represents in traditional Jupyter Widgets

Awesome work @manzt !

benbovy:
Just wanted to report here that I've been experimenting with this PR and it is working great so far!

manzt:
That's great to hear!

I’m curious about what you think of allowing (encouraging?) global (module-level state) being shared between the `instantiate` and `render` hooks. We can do this because we re-load the ESM for each widget instance, so each instance is its own isolated module.

An alternative would be to have a function (like Rollup/Vite plugins) which can create some initial isolated state.

```javascript
export default function () {
  let foo;
  return {
    instantiate({ model }) {
      foo = "bar";
    },
    render({ model, el }) {
      console.log(foo);
    },
  }
}
```

Although I like the named exports, I think this would give us more flexibility going forward to change how the ESM is loaded by the anywidget "runtime".

benbovy:
Hmm I like the isolated state option but I'm afraid I'm not familiar enough with the front-end JS ecosystem to have a proper opinion.

How this approach vs. global module-level state would work with the multiple widgets per ESM workaround described in https://github.com/manzt/anywidget/discussions/382#discussioncomment-7647610?

manzt:
I suppose you could delegate the widget from `initialize` in the default export: 

```javascript
function tableWidget() { /* ... */ }

function chartWidget() { /* ... */ }

const widgetRegistry = { tableWidget, chartWidget };

export default function () {
  let widget;
  return {
    initialize({ model }) {
      const createWidget = widgetRegistry[model.get("component")];
      widget = createWidget();
      return widget.initialize({ model });
    },
    render({ model, el }) {
      return widget.render({ model, el });
    }
  }
}
```

keller-mark:
I prefer the simplicity of the named `render` and `initialize` functions (especially for users who are unfamiliar to JS, they may get tripped up by the need to define a function that exports an object; there are many curly brackets to keep track of). I would imagine that users will copy/paste the boilerplate from the README/docs and modify since it may be tricky to remember.

I am not sure I see the benefit of defining shared state within vs. outside of the function - does it have to do with HMR?
Would this not work?
```js
function tableWidget() { /* ... */ }

function chartWidget() { /* ... */ }

const widgetRegistry = { tableWidget, chartWidget };

let widget;

export function initialize({ model }) {
  const createWidget = widgetRegistry[model.get("component")];
  widget = createWidget();
  return widget.initialize({ model });
}
export function render({ model, el }) {
  return widget.render({ model, el });
}
```

I would not be opposed to the default function export way if there are benefits. It would probably be confusing to support both options (named function exports and default function that returns an object), so if the default function export way provides benefits I would go all-in on that mechanism rather than supporting both.

manzt:
Thanks for the feedback everyone! I plan to merge this week, just need to write up some release notes, documentation, and consider how to deprecate the `render` API.

> I am not sure I see the benefit of defining shared state within vs. outside of the function - does it have to do with HMR?

Currently there isn't really a clear benefit to either choice. However, the function would be a future design consideration, which I think would allow us to avoid potential future breaking changes.

For example, the anywidget "runtime" currently imports a fresh, isolated module per widget instance (i.e., Python class instantiation). This means that things like importing dependencies is duplicated each time a widget in instantiated. 

Having the shared state within the function would allow us experiment with changing the runtime, and avoid breaking changes for end users. For example, we could execute this top-level logic _once_ per widget defintion, meaning that the ESM could define local state that could be shared across all instances. 

With top-level await, this could include additional setup/logic that might be useful:

(**This code is not real and doesn't work**)

```javascript
import * as d3 from "https://esm.sh/d3";

const globalSharedData = await fetch('https://jsonplaceholder.typicode.com/todos/1').then(res => res.json());
const globalSharedState = {};

export default function () {
  const localState = {};
  return {
    instantiate({ model }) {},
    render({ model, el }) {},
}
```

manzt:
Thanks for all the feedback here! I have updated the PR description and implementation. Will merge and then move to updating the documentation for the preferred API.

manzt:
Want to ping some of the existing widget implementations with anticipated deprecation notice

cc: @domoritz @juba @jonmmease 

> [!WARNING]  
> This PR adds a deprecation notice for existing named `render` exports (i.e., all existing widget implementations). I'm not sure we will deprecate in the future, but migrating away from this would allow us more flexibility in loading widget ESM in the future. Migration:
>
> Remove
> ```
> export function render({ model, el }) { ... }
> ^^^^^^
> ```
>  Use
> ```
> function render({ model, el }) { ... }
>          ^^^^^^
> export default { render }
>                  ^^^^^^
> ```

domoritz:
Thanks for the ping. I'll update mosaic.

Judekeyser:
Hello, Just to confirm, what is the least minimal version where this change is not breaking existing code? 
In other words: if I migrate right now my widgets to this newer API, what is the least version I can claim to support (assuming I only use this feature, for simplicity) ?

Thank you in advance!

manzt:
> what is the least minimal version where this change is not breaking existing code?

Great question!

The API is of [`0.9.0`](https://github.com/manzt/anywidget/releases/tag/anywidget%400.9.0). You can pin `anywidget>=0.9.0` in your `pyproject.toml` etc.

flekschas:
> The API is of [0.9.0](https://github.com/manzt/anywidget/releases/tag/anywidget%400.9.0). You can pin anywidget>=0.9.0 in your pyproject.toml etc.

@manzt If possible, it'd be nice to include this in the deprecation warning :)

Issue #374, opened 2023-11-10
`(core dumped) poetry run pytest` on CI when `_should_start_thread` is `True`


kylebarron:
I'm updating anywidget to version 0.7.1 to take advantage of the really nice https://github.com/manzt/anywidget/pull/345. One consequence is that I can't run CI tests without the JS parts built, because if those files aren't found, anywidget will error.

I've been hitting a very consistent `(core dumped) poetry run pytest` on Github Actions CI when the JS files exist and when `watchfiles` is installed. It looks to be some sort of threading issue, and so not having `watchfiles` installed ensures that `_should_start_thread` is `False`:
https://github.com/manzt/anywidget/blob/2600bd8a85c7e8e5f5ce572410592314b713ec72/anywidget/_util.py#L178-L187

Here's a minimal example of what I'm seeing: https://github.com/developmentseed/lonboard/pull/234

```
Run poetry run pytest
============================= test session starts ==============================
platform linux -- Python 3.8.18, pytest-7.4.3, pluggy-1.3.0
rootdir: /home/runner/work/lonboard/lonboard
plugins: anyio-4.0.0
collected 11 items

tests/test_layer.py ..                                                   [ [18](https://github.com/developmentseed/lonboard/actions/runs/6827322556/job/18569190597#step:13:19)%]
tests/test_map.py .                                                      [ 27%]
tests/test_traits.py ........                                            [100%]

============================== 11 passed in 3.66s ==============================
FATAL: exception not rethrown
/home/runner/work/_temp/4f63a42f-5562-4a63-a9ca-331e37c02752.sh: line 1:  [22](https://github.com/developmentseed/lonboard/actions/runs/6827322556/job/18569190597#step:13:23)85 Aborted                 (core dumped) poetry run pytest
```

This is using anywidget `0.7.1` and watchfiles `0.21.0`

manzt:
maybe we could create a environment variable? to force no HMR?

kylebarron:
Yeah, that would be an ok solution... As a workaround I make sure I don't install `watchfiles` on CI. It would be _nice_ if we could figure out what's actually wrong, but that might be a 🕳️

manzt:
Hmm, ok looking into this. The root of the issue is that we automatically read the file if a path is detected in class definition. Maybe we should delay this until the first instance of the class is created.

kylebarron:
[...]

PR #424, opened 2024-01-31
feat(anywidget): Make HMR explicit and opt-in

### LINKED ISSUE #423: Shiny live altair example not working
https://shinylive.io/py/examples/#altair

```
>>> Installing altair ...
Installing anywidget ...
Installing palmerpenguins ...
Unhandled error: can't start new thread
```

I believe this is related to the issue with https://github.com/manzt/anywidget/issues/374.

It seems that starting a watchfiles thread if a ESM/CSS sources are outside of `site-packages` leads to a bunch of difficult issues to track down in various environments using anywidget. This should only be happening during development, and I'm starting to think explicitly enabled by developers. That way anywidget doesn't try to guess what it should do (and fail for situations it's not aware of).

The alternative would be to try to handle these edge cases, but that seems untenable.

manzt:
Fixes #423 #374 

## What

Makes anywidget's native HMR explicit and opt-in via the `ANYWIDGET_HMR` environment variable. 

## Context

The built-in HMR is only meant to be used during "live" development, when _local_ widgets are in imported and used in an active notebook. However, our heuristics for enabling this development feature lead to many false positives, and thus issues  issues in various non- "live" development settings #423, #374, #276.

**This is a bad experience for end users**. 

It doesn't seem possible to account for all these edge cases, and since HMR is meant as a development only feature, it makes sense to make it explicit and opt-in _by the developer_. This PR requires an environment variable to apply anywidget's prior heuristics. It can be set in a notebook:

```python
%env ANYWIDGET_HMR=1
```

or when launching a notebook from the command line:

```python
ANYWIDGET_HMR=1 jupyter lab
```

### DIFF
--- a/.changeset/shiny-houses-run.md
+++ b/.changeset/shiny-houses-run.md
@@ -1,0 +1,5 @@
+---
+"anywidget": minor
+---
+
+feat: Require explicit opt-in to HMR during development
--- a/.changeset/shiny-houses-run.md
+++ b/.changeset/shiny-houses-run.md
@@ -2,4 +2,4 @@
 "anywidget": minor
 ---
 
-feat: Require explicit opt-in to HMR during development
+feat: Require `ANYWIDGET_HMR` to opt-in to HMR during development
--- a/anywidget/_util.py
+++ b/anywidget/_util.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import os
 import pathlib
 import re
 import sys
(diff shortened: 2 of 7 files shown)


nvictus:
As a sage once said:

```python
>>> import this
The Zen of Python, by Tim Peters

Beautiful is better than ugly.
Explicit is better than implicit.
...
```

nvictus:
Would it also make sense to also provide an explicit hook to activate HMR at runtime in the notebook?

manzt:
> Would it also make sense to also provide an explicit hook to activate HMR at runtime in the notebook?

We could definitely add an API later for that, but you can currently set within a notebook with:

```
%env ANYWIDGET_HMR=1
```

PR #437, opened 2024-02-06
feat(create-anywidget): Add "readme" to pyproject.toml template

kolibril13:
This pr is adding readme = "README.md" 
This way e.g.

image

will become

image

### DIFF
--- a/.changeset/afraid-panthers-explain.md
+++ b/.changeset/afraid-panthers-explain.md
@@ -1,0 +1,5 @@
+---
+"create-anywidget": patch
+---
+
+Add README.md to pyproject.toml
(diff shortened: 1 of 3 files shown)


manzt:
Thanks! Just changed the name of the PR because I want to avoid using the name "cookiecutter" for our CLI. It doesn't use [cookiecutter](https://cookiecutter.readthedocs.io/en/stable/) and helps destinguish our CLI fro the existing cookiecutters from Jupyter.

manzt (changes requested):
Looks great! Thanks. CI is failing because we have a test to generate some repos using the CLI and take a "snapshot" of the output files. 

The line added in this PR now is in each snapshot (good), causing it to fail, so we need to update the snapshot. Please:

```sh
cd packages/create-anywidget
```

```sh
pnpm vitest --update # updates snapshot
```

And make sure the diff looks correct. Check in the new snapshots.

Then go to the repo root:

```sh
cd ../../
```

and create a changeset for create-anywidget. 

```sh
pnpm changeset # follow prompt
```

manzt:
Don't want this to go stale. In the future, please follow the steps outlined above for making a contribution for `create-anywidget`.

kolibril13:
hi @manzt,
thanks for updating the snapshot and for merging the pr!

PR #438, opened 2024-02-10
refactor(anywidget): Use signals to track runtime state

manzt:
Refactors the anywidget runtime to use signals to track dependencies. This cleans up the implementation quite a bit, managing updates as effects. Previously there were duplicated calls to the cleanup functions returned from each of the lifecycle hooks.


manzt:
See how we correctly cleanup and dispose of views and initialize:


https://github.com/manzt/anywidget/assets/24403730/f3ef10ef-d856-4ad9-ae24-e6419a1b5d5c

PR #161, opened 2023-06-22
feat!: Drop support for Python 3.7

manzt:
Think I'm doing this right, lol cc: @tlambert03. Next release will be a minor release.

- feat: Drop support for Python 3.7
- chore: update build-system to use jlab v4

PR #441, opened 2024-02-10
feat!: Add support for Python 3.12 and drop Python 3.7

manzt:
Related 
- https://github.com/manzt/anywidget/pull/161
- https://github.com/manzt/anywidget/issues/166
- https://github.com/manzt/anywidget/pull/167

We've tried to keep backward compatibility with Python 3.7, but now it is getting in the way of supporting 3.12. Python 3.7 has been EOL for almost a year, and many others in the ecosystem have moved on (including JupyterLab). If we were able, I'd try to keep support but it's causing too many issues with latest version of Python.

Let me explain. Custom Jupyter Widgets (which anywidget is internally) requires building with JupyterLab (via `@jupyterlab/builder`). In order to build a custom widget, and hence anywidget from scratch the build system must run:

```sh
jupyter labextension build
# - Requires JupyterLab
#    - Requires `@jupyterlab/builder`
```

The version of JLab (Python) and `@jupyterlab/builder` are tied deeply tied together. This is a problem. It has meant that in order to keep support for Python 3.7, we've needed to keep our build system on JupyterLab 3 and an older version of `@jupyterlab/builder`, since JupyterLab 4 has dropped support for Python 3.7. This older version of JupyterLab does not work on Python 3.12 (issue with pyzmq), which means that you need to have two different versions of Python to build and run anywidget (**bad**).

anywidget's philosophy is to be low friction, while pushing the ecosystem forward. If we were able to move off of JupyterLab as a build system dep, it may be possible to support this legacy use case, but this doesn't not seem possible without extreme build-tooling hell.

Ecosystem:

- Python 3.7 is EOL
- Google Colab is Python 3.10
- JupyterLab 4 has dropped support for Python 3.7
- `ipywidgets` pyproject.toml says 3.7 support, but [tests are only run on >=3.8](https://github.com/jupyter-widgets/ipywidgets/blob/b78de43e12ff26e4aa16e6e4c6844a7c82a8ee1c/.github/workflows/tests.yml#L80)

### DIFF
--- a/.changeset/clever-houses-float.md
+++ b/.changeset/clever-houses-float.md
@@ -1,0 +1,5 @@
+---
+"anywidget": minor
+---
+
+Drop support for Python 3.7 and add Python 3.12
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]
+        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12"]
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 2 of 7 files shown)


manzt:
pyzmq doesn't seem to build for 3.12?

PR #451, opened 2024-02-23
Add files for a JupyterLite demo

jtpio:
Opening this PR early (still as a draft) to see if there would be an interest to offer a JupyterLite demo, so users can try `anywidget` in the browser easily.

This demo is based on `jupyterlite-xeus` (https://jupyterlite-xeus.readthedocs.io) and installs `anywidget` from conda-forge, so the example notebook works without additional code.

For now it likely only support inline `_esm` and `_css`, but should let users author a simple widget in their browser. Maybe there could then be a way to to "mount" additional files (such as `index.js` and `index.css`) so they are editable in JupyterLab, and correctly picked up by `anywidget`.

Here is a demo screencast:

[anywidget-jupyterlite-demo.webm](https://github.com/manzt/anywidget/assets/591645/b1f60976-e936-48f9-a100-2513d5ef0d87)

If there is interest, I can look into integrating this with the existing documentation (maybe even inline a notebook in the docs somewhere).

### TODO

- [x] Environment files to create the JupyterLite demo
- [ ] Integrate with the docs
- [ ] Contributing docs


jtpio:
Actually this can already easily be tested on any JupyterLite deployment that has `anywidget` installed as a build dependency (to properly pick up the lab extension). For example here: https://jtp.io/lit/lab/index.html?path=pyodide/anywidget.ipynb

image

manzt:
Wow, this is fantastic! I looked into setting up a JupyterLite notebook a while back but ran into some hiccups getting the frontend JS for anywidget to be discovered by the kernel. I'm guessing you need a preconfigured JupyterLite instance with this enabled. 

Overall, I'd love to build "the notebook" for teaching widgets (via anywidget). It would be awesome to have this live in a single Jupyter notebook that we could link folks to to learn about specific widget concepts. In theory, we could refer to this rather than "Jupyter Widgets: the good parts" section of the documentation.

jtpio:
> I'm guessing you need a preconfigured JupyterLite instance with this enabled.

Yes, the `anywidget` Python package needs to be installed in the build environment so its JupyterLab extension can be found at build time.

Playing a bit more with it, it actually supports reading files that can be edited from the UI directly (but no HMR): 

[anywidget-jupyterlite-files.webm](https://github.com/manzt/anywidget/assets/591645/f13c7a13-4c08-4c10-848f-7023b4221501)

jtpio:
Indeed this would be a low-friction way for trying something out quickly!

@manzt how do you see something like this integrated in the repo?

I quickly looked into adding this to the Astro docs site:

image

Although this would require installing the `jupyterlite` CLI, `anywidget` and a few other Python packages to be able to index files and potentially enable more features.

This could complicate the Netlify build and setup. Maybe a first step could be to deploy this JupyterLite demo to GitHub pages, and link to it from anywidget.dev?

jtpio:
> This could complicate the Netlify build and setup. Maybe a first step could be to deploy this JupyterLite demo to GitHub pages, and link to it from anywidget.dev?

For reference it's possible to run the `jupyter lite build` command on hosts like Vercel and Netlify (https://jupyterlite.readthedocs.io/en/latest/howto/deployment/vercel-netlify.html). Although it's usually better to use `micromamba`, `pixi` or maybe `hatch shell` to create a virtual environment on these hosts during the build step, to not interfere with their Python setup.

jtpio:
> Maybe a first step could be to deploy this JupyterLite demo to GitHub pages

Just put up a self-contained demo here: https://github.com/jtpio/anywidget-lite

manzt:
This is great. I have kind of a hectic week so I might be a bit blocked on thinking about integrating into the documentation. However, I really like the idea. We have a "notebook" section to the docs, which prerenders notebooks in the astro site. I'd probably be in favor of moving that section to embedded JupyterLite.

PR #453, opened 2024-02-24
feat(experimental): Add `command` system for calling Python from JS

manzt:
This PR introduces an experimental API designed to make sending custom messages easier in the front end. This is achieved through a new RPC mechanism, which allows the frontend to `invoke` a `command` registered in Python and await the response.

A minimal example:

```py
import anywidget
import traitlets

    _esm = """
    async function render({ model, el, experimental }) {
      let [msg, buffers] = await experimental.invoke("_echo", "hello");
      console.log(msg) // "hello"
    }
    export default { render };
    """
    
    @anywidget.experimental.command
    def _echo(self, msg, buffers):
        return msg, buffers
```

this can make implementing things with dynamic data much easier, like the `ZarrWidget`:

```py
import zarr
import zarr.storage

store = zarr.storage.MemoryStore()
grp = zarr.open(store)
grp.create_dataset("foo", shape=[10, 10])

ZarrWidget(store=store)
```

<details>
<summary> <code>ZarrWidget</code> implementation </summary>

```py
import anywidget
import traitlets

class ZarrWidget(anywidget.AnyWidget):
    _esm = """
    import * as zarr from "https://esm.sh/zarrita@next";
    
    function render({ model, el, experimental }) {
      let root = zarr.root({
        async get(key) {
          const [data, buffers] = await experimental.invoke("_zarr_get", key);
          if (!data.success) return undefined;
          return buffers[0].buffer;
        },
      });
    
      let pre = document.createElement("pre");

      btn.innerHTML = `click to load`;
      btn.addEventListener("click", async () => {
        const node = await zarr.open(root.resolve("/foo"), { kind: "array" });
        const { data, shape, stride } = await zarr.get(node);
        pre.innerText = JSON.stringify({ data: data.constructor.name, shape, stride });
      });
      el.append(btn, pre);
    }
    export default { render };
    """
    def __init__(self, store, **kwargs):
        super().__init__(**kwargs)
        self._store = store

    @anywidget.experimental.command
    def _zarr_get(self, key, buffers):
        try:
             buffers = [store[key.lstrip("/")]]
        except KeyError:
             buffers = []
        return { "success": len(buffers) == 1 }, buffers
```

</details>

cc: @keller-mark 

TODO:

- [x] Add tests
- [x] Support sending binary data from `invoke` (via another argument)
- [ ] Documentation

Another option would be to move this kind of dispatching mechanism to a separate library (e.g., `@anywidget/std`), but, since it is coupled to the Python, I prefer we implement something within anywidget that makes this pattern frictionless for applications that need it.

EDIT: This PR has changed from a messaging callback to registering methods like RPC.


keller-mark:
This is awesome! Very nice how the `on_msg` only gets called if the user has defined the `_anywidget_experimental_reducer`

manzt:
> Very nice how the `on_msg` only gets called if the user has defined the `_anywidget_experimental_reducer`

Thanks! Yeah, I think having something opinionated but opt-in would probably be preferable. I renamed to `_experimental_anywidget_reducer` in anticipation that we would remove the leading `_experimental` when stabilized.

kylebarron (approved):

kylebarron:
I'm still curious whether the Python callback can be made async or whether that messes with the Jupyter event loop

anywidget/experimental.py
```diff
@@ -103,3 +106,31 @@ def _decorator(cls: T) -> T:
...
+        return
+
+    def handle_anywidget_dispatch(
+        self: AnywidgetReducerProtocol, msg: str | list | dict, buffers: list[bytes]
+    ) -> None:
+        if not isinstance(msg, dict) or msg.get("kind") != "anywidget-dispatch":
+            return
+        response, buffers = widget._experimental_anywidget_reducer(
+            msg["action"], buffers
+        )
```
kylebarron, line 126: I'd suggest for this to be in a try/except. I think if an error is raised in the user callback, then the error gets swallowed somewhere. Maybe on error you could propagate something back to the JS so you're not blocking the Python kernel any more?

packages/anywidget/src/widget.js
```diff
@@ -242,6 +242,35 @@ function throw_anywidget_error(source) {
...
 }
 
+/**
+ * @template T
+ * @param {import("@anywidget/types").AnyModel} model
+ * @param {any} [action]
+ * @param {{ timeout?: number }} [options]
+ */
+export function dispatch(model, action, { timeout = 3000 } = {}) {
+	let id = Date.now().toString(36);
```
kylebarron, line 252: I found that you really need a uuid here. In lonboard, the deck TileLayer can make multiple requests at the same time and this time-based ID isn't stable enough. Switching to `uuid`'s `v4()` fixed my issues.

manzt: ok great to know. i will switch

manzt:
I've been reviewing this, and I'm realizing this doesn't really follow the redux pattern. In Redux, `dispatch`ing emits and action, which updates the state, and then view updates according to state changes. Therefore, `model.send` is much closer conceptually to Redux-y patterns (i.e., emit a message to Python, Python updates the model, JS view responds to changes on the model).

This PR is narrowing in on something different but very important: RPC. So I wonder about taking this one step further, and baking in a mechanism like [Tauri's `invoke`](https://tauri.app/v1/guides/features/command/). Essentially, I wonder about something like:


```py
    _esm = """
    async function render({ model, el, experimental }) {
      let [response, buffers] = await experimental.invoke("_echo", "hello, world");
      // Handle the response
      console.log(response) // pong
    }
    """
   
   @anywidget.command
   def _echo(msg, buffers):
       return msg, buffers
```

I think this is the direction I'd like to take this PR, what do you think @kylebarron @keller-mark ?

anywidget/experimental.py
manzt: Ah, that's a great idea! We could throw a JS error for a special message sent back from Python.

keller-mark:
I think framing it more like RPC conceptually makes sense. I do not have a super strong preference towards `dispatch` or `invoke` as long as it allows the ZarrWidget to be implemented.

Is the `@anywidget.command` decorator necessary? Maybe it helps with security or future-proof things to know which functions are potentially `invoke`d?

> It would be nice if we could have something other than msg/buffers, I suppose we could require "msg" to be an dict and **msg when invoking the callback... but then not sure what to do about buffers.

I don't really follow this point but i am not familiar with `msgpack`. Is the issue that you want the API to be 

```diff
-let [msg] = await experimental.invoke("_echo", "hello, world");
+let msg = await experimental.invoke("_echo", "hello, world");
```

when the user does not care about buffers?

manzt:
> do not have a super strong preference towards dispatch or invoke as long as it allows the `ZarrWidget` to be implemented.

Right, makes sense. I guess I should just stop bike-shedding and merge the thing :) 

> Is the `@anywidget.command` decorator necessary?

This decorator "marks" the methods on the class to "expose" to the front end. I guess we could try to inspect the signature of all the methods on the class and try to guess which ones implement the invoke signature:

```py
(msg: Any, buffers: list[bytes]) -> tuple[Any, list[bytes]]
```

but this is more explicit and we can raise type errors if the wrapped method has the wrong signature.

> I don't really follow this point but i am not familiar with `msgpack`. Is the issue that you want the API to be.

If you wanted to return a `dict` of buffers currently (for example), using `getitems` from zarr-python:

```py

def getitems(keys: list[str]) -> dict[str, bytes] ... 
```

You would need to extract the buffers manually, both on the JS and Python side:

```py
@anywidget.command
def _getitems(self, msg: Any, buffers: list[bytes]) -> tuple[list[str], list[bytes]]:
    keys = msg["keys"]
    items = self.store.getitems(keys)
    return list(items.keys()), list(items.values())
```

If this was a traitlet, Jupyter Widgets would automatically take care of "extracting" the buffers _and_ assembling the deconstructed object on the JS side. But for custom messages, you need to do it yourself. I guess I was wondering if this is a place we could further make developing widgets easier, so developers didn't need to worry about packing/unpacking buffers.

One option would be to reuse the packing mechanism from Jupyter Widgets, or a more well defined binary protocol like msgpack. My current thinking is that we could extend `command` to allow this as an option (not for this PR):

```py
@anywidget.command(serialize=None)
def _getitems(self, msg: Any, buffers: list[bytes]) -> tuple[list[str], list[bytes]]:
    keys = msg["keys"]
    items = self.store.getitems(keys)
    return list(items.keys()), list(items.values())
```

```py
@anywidget.command(serialize="msgpack") # serialize="jupyter-widgets"
def _getitems(self, keys: list[str]) -> dict[str, bytes]:
    keys = msg["keys"]
    return self.store.getitems(keys)
```

kylebarron:
I agree that an RPC-based API makes sense for this, and also +1 on exploring msgpack (at least as an option). I haven't used it myself but I'm guessing the client libraries for that are pretty stable and minimal

manzt:
Awesome. I'm going to push forward and hopefully cut a release this week!

manzt:
> I haven't used it myself but I'm guessing the client libraries for that are pretty stable and minimal

Yes, I'm familiar with msgpack for neovim's extension mechanism. On the Python side there is [`msgpec`](https://jcristharif.com/msgspec/), which is minimal and speedy, and on the JS side [`@msgpack/msgpack`](https://www.npmjs.com/package/@msgpack/msgpack).

keller-mark:
Thanks for the clarification, the ability to specify in the decorator whether the msgpack serialization should be used looks like a great way forward!

manzt:
Cool, let's merge the thing! I will work on some docs in a separate PR, but they will be a draft likely until we stabilize the feature.

cc: @mscolnick, just FYI a new API we are iterating on. It would be great to learn if this is something that you think Marimo could support. Relevant code:

https://github.com/manzt/anywidget/blob/568e2431c5f9a3644c32696ff9e0a97aa8f59628/packages/anywidget/src/widget.js#L241-L283

mscolnick:
@manzt very cool addition! we can support this (we have a concept of RPC/Functions too that I can adapt to work with this as well. 

Is there a timeline for when it will be out of experimental? Would like to avoid 2 changes if possible, but not against it.

abel1502:
Is there a way to redirect the command's output into an `ipywidgets.Output`? (Or is its output visible automatically?). I feel that would be almost a necessity for debugging commands

PR #472, opened 2024-03-07
docs: edit README in create-anywidget

ManonMarchand:
Hi here, 

As discussed on discord, here are instructions for the development installation for create-anywidget.

I did not deviate a lot from your suggestion :slightly_smiling_face:


ManonMarchand:
Should be good?

Issue #475, opened 2024-03-08
create-anywidget: order of "Next steps"


dvdkouril:
Hi there,
this is a small thing, and could be a matter of taste (or something else I'm not considering). Feel free to close.

With the create-anywidget script, you get the following instructions at the end:
```
Next steps:
  1: cd test
  2: git init && git add -A && git commit -m "Initial commit" (optional)
  3: npm install
  4: npm run dev
```
This means that you make the initial commit before running `npm install`. `npm install` then produces `package-lock.json`. AFAIK, it's good practice to push lockfiles to origin, so you need to make an extra second commit.

Comparing to `npm create svelte@latest`:
<img width="597" alt="Screenshot 2024-03-08 at 3 07 24 PM" src="https://github.com/manzt/anywidget/assets/2307442/ffd3e049-b5de-4420-89ad-8495f03840a2">

I can take the bullet and make this complex change, if you agree. But then again, just close if there's a reason it's the way it currently is.

manzt:
Nope no reason! I think I actually copied from Sveltekit, so they must have changed this recently :)

manzt:
Fixed in #476, thanks!

PR #476, opened 2024-03-08
Change order of "Next steps" in create-anywidget

dvdkouril:
Reasoning in #475

### DIFF
--- a/.changeset/eleven-readers-wash.md
+++ b/.changeset/eleven-readers-wash.md
@@ -2,4 +2,4 @@
 "create-anywidget": patch
 ---
 
-Change order of "Next steps" after running create-anywidget script (new: npm install before git commands)
+Change order of "Next steps" (npm install before git commands)
--- a/.changeset/eleven-readers-wash.md
+++ b/.changeset/eleven-readers-wash.md
@@ -1,0 +1,5 @@
+---
+"create-anywidget": patch
+---
+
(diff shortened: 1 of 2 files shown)


manzt:
Thanks for the contribution! The error is coming from some tests i have to make sure the exact output of the cli stays the same. You can update the snapshots with pnpm vitest and u to update the snapshot. Otherwise i can take care of it later :)

dvdkouril:
Hm, I thought the error's coming from some linting line length rule applied on the changeset.
Anyway, I also realized, that it probably makes sense to have it:
1. `cd`
2. `npm install`
3. `git stuff`
4. `npm run dev`
5. `"To close..."`

While what I committed gives:
1. `cd` 
2. `npm install`
3. `npm run dev`
4. `"To close..."`
5. `git stuff`

and `npm run dev` hijacks your shells so you don't even get to do your commit.

manzt:
> Hm, I thought the error's coming from some linting line length rule applied on the changeset.

Ah ya, you are right. Just need to run `pnpm format`.

> Anyway, I also realized, that it probably makes sense to have it

Yah, I like this option, running `npm run dev` last (after optional git stuff).

manzt:
@dvdkouril I took a stab at your suggested changes. Look ok?

dvdkouril:
Yes! So sorry I let this one rot. I was thinking whether there's a way to write it without repeating the if loop.
But this looks good to me.

PR #505, opened 2024-03-25
docs: Add sitemap

manzt:
cc: @Coderambling


Coderambling:
Wow that was quick! Was the sitemap itself generated by the Astro template( and what framework is that a part of)?

Does this mean the sitemap will be updated when the URL structure of the site changes?

So when https://anywidget.dev/en/community/ is published on the sitewith the showcases, will that URL be automatically added to the sitemap.xml?

manzt:
The docs are an [Astro](https://astro.build) site. Yes, the site map is generated at build time with that package. There are more details on how to configure it in [their docs](https://docs.astro.build/en/guides/integrations-guide/sitemap/).

PR #510, opened 2024-03-29
fix(docs): Fix typo in vite config

donmccurdy:
Small typo, `plugin` -> `plugins`: https://vitejs.dev/config/shared-options.html#plugins


manzt (approved):
thanks!

Issue #385, opened 2023-11-20
importing a js library fails in Jupyter Hub, works in Jupyter Lab


mlamoureux:
I'm creating an "anywidget" to access some hardware devices known as "Phidgets."
This requires importing a library of js code in the _esm file for the anywidget.
My import statement works fine in Jupyter Lab, but not in Jupyter Hub (the classic Jupyter notebook). 
My question is: what is the proper way to import a js module into the _esm file?

This statement works in Jupyter Lab, but not in Hub (i.e. not in classic notebook):
    _esm = """
      import  "https://unpkg.com/phidget22/browser/phidget22";
     ....

I've also tried the following, based on the examples in the anywidget docs:
import * as phidget22 from "https://unpkg.com/phidget22/browser/phidget22";  or
import USBConnection from "https://unpkg.com/phidget22/browser/phidget22";

But they don't work at all. So, what is the right way to import?

manzt:
> My import statement works fine in Jupyter Lab, but not in Jupyter Hub (the classic Jupyter notebook).

Thank you for raising this issue. The problem you're encountering likely stems from the way Jupyter Widgets are installed and recognized within the Jupyter ecosystem, rather than an issue with your code.

Jupyter Widgets need to be detected by Jupyter upon kernel startup, meaning you have to restart the kernel to load new widgets. This is a known limitation, distinct from how regular Python libraries are installed. For regular libraries, installation in user-land suffices, but widgets need to be installed in Jupyter Hub itself.

I think I had with @hamelin at SciPy earlier this year about this very issue with classic Jupyter Widgets. Since anywidget is a classic widget, it must be installed in Jupyter Hub for all users. However (I believe) once anywidget is installed successfully, any custom widget developed using anywidget can be installed and managed in user-land.

To resolve this, you need to install anywidget into Jupyter Hub. This should hopefully enable the proper functioning of your import statement in Jupyter Hub.

- https://github.com/jupyter-widgets/ipywidgets/issues/1949

manzt:
There might be two issues here:

1.) The difference between JupyterLab and Jupyter Hub for importing `anywidget` (outlined above)
2.) "[unpkg.com/phidget22/browser/phidget22](https://unpkg.com/phidget22/browser/phidget22)" is not an ECMAScript Module (ESM)

[...]

PR #521, opened 2024-04-02
Allow semver-compatible frontend versions

kylebarron:
In relation to the discussion about JupyterHub in https://github.com/manzt/anywidget/issues/385, I've suggested to a few of my users to install anywidget on the server, then install lonboard in the client environment. Lonboard does not pin a version of anywidget, so when a new anywidget version is released, the next install of lonboard will break because the client anywidget version does not exactly match the server anywidget version.

I have reason to believe that using `"~major.minor.*"` works, because pydeck has been doing that for its frontend code for a while, to allow the Python code and JS bindings to be versioned separately (see [1](https://github.com/visgl/deck.gl/blob/9117414ef894b8755e0eff33e7cf503a2137249f/bindings/pydeck/pydeck/frontend_semver.py#L1), [2](https://github.com/visgl/deck.gl/blob/ff7566644bc716cf917c0f21e8155b5ff1074b7a/bindings/pydeck/pydeck/widget/_frontend.py#L14), [3](https://github.com/visgl/deck.gl/blob/9117414ef894b8755e0eff33e7cf503a2137249f/bindings/pydeck/pydeck/widget/widget.py#L55)). 

That said, I haven't yet tested this in a JupyterHub environment. (I haven't tried to set up the development environment yet)

### DIFF
--- a/.changeset/cuddly-fishes-heal.md
+++ b/.changeset/cuddly-fishes-heal.md
@@ -1,0 +1,7 @@
+---
+"anywidget": patch
+---
+
+Relax version pinning for anywidget front end
+
+Adopted `~major.minor.*` notation for more flexible version compatibility in the front end, mirroring practices improve compatability in environments where bumping the front-end versions is not possible for end users (i.e., JupyterHub). This change is intended to enhance adaptability without causing disruptions. If issues arise, please report them on our issues page.
--- a/anywidget/_descriptor.py
+++ b/anywidget/_descriptor.py
@@ -83,7 +83,7 @@ def get_semver_version(version: str):
     if is_pre_release:
         return ".".join(split)
 
-    return "~" + ".".join([split[0], split[1], "0"])
+    return "~" + ".".join([split[0], split[1], "*"])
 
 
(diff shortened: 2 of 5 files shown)


manzt:
Hey, just wanted to check in on this. I notice the PR is still in a Draft state. I should have some time to push it forward if that is ok with you. Thanks!

kylebarron:
I only set it as "draft" because I hadn't figured out how to test it yet. But it "looks" valid based on what pydeck was using

manzt:
Ah, sorry for the misunderstanding. I thought it wasn't ready for review. I'll have a look and get back to you :) thanks for the contribution!

manzt:
Ok, I tried this out in a few different environments (Lab, standalone HTML, Colab) and things seem to be working. The various CDN importers seem to resolve the version specifier correctly, which would be the main thing I would worry about. 

I think this should be seen as a temporary fix. I still think we should ideally separately version the anywidget frontend in the future.

kylebarron:
awesome 🙌

PR #526, opened 2024-04-03
Refactor AnyWidget command registration

tlambert03:
separates the collection and registration of anywidget commands.  `_collect_anywidget_commands` happens on `__init_subclass__` ... and doesn't require getting any attributes on widget instances.  while `_register_anywidget_commands` still happens on widget init.

just an option. not necessarily much better :)

### DIFF
--- a/.changeset/ten-apricots-float.md
+++ b/.changeset/ten-apricots-float.md
@@ -1,0 +1,5 @@
+---
+"anywidget": patch
+---
+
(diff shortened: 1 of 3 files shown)


manzt:
Ooo excited to have a look, thanks :) You can ignore the JS tests, i need to update some snapshots (they are broken on main).

manzt (approved):
I think I like this better, thanks a lot for the help!

PR #536, opened 2024-04-11
docs: cleanup after requestAnimationFrame example

dvdkouril:
Quick sketch of the scenario where you use the cleanup function returned from `render()` to call `cancelAnimationFrame` to break the continuously updating frameloop.

I still want to think about whether this is the most minimal and expressive code example (that's why I'm keeping it a draft). There also might be a better way to word the description.

### DIFF
--- a/docs/src/pages/blog/anywidget-02.md
+++ b/docs/src/pages/blog/anywidget-02.md
@@ -219,28 +219,19 @@ Another common scenario requiring proper cleanup is when using
 `requestAnimationFrame` to continually redraw into a canvas:
 
 ```javascript
-class MyCustomRenderer {
-	constructor() {
-		this.render = this.render.bind(this);
-		this.redrawRequest = requestAnimationFrame(this.render);
-		this.canvasEl = document.createElement("canvas");
-	}
-
-	render() {
-		this.redrawRequest = requestAnimationFrame(this.render);
-
-		// ...custom rendering goes here...
-	}
-}
-
 export default {
 	render({ model, el }) {
-		const renderer = new MyCustomRenderer();
-
-		el.appendChild(renderer.canvasEl);
+		const canvasEl = document.createElement("canvas");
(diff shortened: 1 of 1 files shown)


docs/src/pages/blog/anywidget-02.md
```diff
@@ -215,6 +215,40 @@ export default { render };
...
+	render({ model, el }) {
+		const renderer = new MyCustomRenderer();
+
+		el.appendChild(renderer.canvasEl);
+
+		return () => {
+			cancelAnimationFrame(renderer.redrawRequest);
+		};
+	},
+};
```
manzt, line 246: Love this! Thanks for the contribution. I think for now, this makes total sense to put in this blog post... but one of these days I'd like to pull out useful nuggets from these for some part of the dedicated documentation.

For this example, the only thing I'm wondering is if we could inline some of this within `render` rather than introducing new syntax like a `class`. It's mostly what would be familiar to an end user. I'm guessing if you're using 	requestAnimationFrame`, a class would probably be familiar .... so maybe i'm over thinking it.

```js
export default {
	render({ model, el }) {
		const canvasEl = document.createElement("canvas");
		let requestId = requestAnimationFrame(step);
		
		function step() {
			/* ...custom rendering goes here... */
			requestId = requestAnimationFrame(step);
		}
		el.appendChild(canvasEl);

		return () => {
			cancelAnimationFrame(requestId);
		};
	},
};
```

Stealing the variable names from [`MDN`](https://developer.mozilla.org/en-US/docs/Web/API/Window/cancelAnimationFrame).

manzt (approved):
Just a small comment. Thanks for the contribution!

docs/src/pages/blog/anywidget-02.md
dvdkouril: Yes, this looks much better, more self-contained! I'll make the change.

manzt (approved):

PR #540, opened 2024-04-18
feat(experimental): Replace invoke timeout with `AbortSignal`

manzt:
This allows more flexible control over aborting the invoke request, including delegating to third-party libraries that manage cancellation.

```js
export default {
  async render({ model, el }) {
    const controller = new AbortController();

    // Randomly abort the request after 1 second
    setTimeout(() => Math.random() < 0.5 && controller.abort(), 1000);

    const signal = controller.signal;
    model
      .invoke("echo", "Hello, world", { signal })
      .then((result) => {
        el.innerHTML = result;
      })
      .catch((err) => {
        el.innerHTML = `Error: ${err.message}`;
      });
  },
};
```

### DIFF
--- a/.changeset/three-snails-give.md
+++ b/.changeset/three-snails-give.md
@@ -1,0 +1,29 @@
+---
+"anywidget": patch
+"@anywidget/types": patch
+---
+
+Experimental: Replace invoke timeout with more flexible `AbortSignal`
+
+This allows more flexible control over aborting the invoke request, including delegating to third-party libraries that manage cancellation.
+
+```js
+export default {
+  async render({ model, el }) {
+    const controller = new AbortController();
+
+    // Randomly abort the request after 1 second
+    setTimeout(() => Math.random() < 0.5 && controller.abort(), 1000);
+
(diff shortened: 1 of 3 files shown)


kylebarron:
😍

manzt:
> I think it might be worth having both a timeout and a signal. Like set timeout to 10 seconds but stop sooner if the signal was aborted?

I guess I'd defer this to the caller with [`AbortSignal.any`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/any_static#using_abortsignal.any). What do you think?

```js

let res = await invoke("_get_tile", tileId, {
  signal: AbortSignal.any([deckSIgnal, AbortSignal.timeout(10_000)]),
});

```

kylebarron:
> I guess I'd defer this to the caller with [`AbortSignal.any`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/any_static#using_abortsignal.any). What do you think?

Oh that's cool! I didn't know that existed. That seems fine!

PR #598, opened 2024-05-27
feat(vite): Support default export of render in Vite HMR wrapper

### LINKED ISSUE #586: Vite HMR expects render function to be exported directly
# Description

When the `render` function of the `_esm` module is exported using `export default  { render }`, Vite HRM fails because it [looks](https://github.com/manzt/anywidget/blob/c037e479cf2ce931ac69422d93a371397fecd37b/packages/vite/index.js#L30) for the `render` function at the root of the module namespace object rather than under the `default` property. Hence a `render is not a function` error ensues.

Additionally, since the HMR wrapper [exports](https://github.com/manzt/anywidget/blob/c037e479cf2ce931ac69422d93a371397fecd37b/packages/vite/index.js#L27) `render` directly, a deprecation warning discouraging this practice is displayed in the console.

# Context

Very much inspired by your work on `pygv`,  I am tinkering with [`pyjb`](https://github.com/ScilifelabDataCentre/pyjb/tree/main), an early attempt to bring [JBrowse](https://jbrowse.org/jb2/) on planet Jupyter, propelled by `anywidget` of course :)  Here is the [`_esm`](https://github.com/ScilifelabDataCentre/pyjb/blob/main/src/index.js) and a [notebook](https://github.com/ScilifelabDataCentre/pyjb/tree/main/examples) to reproduce the issue.

# Workaround 

A direct export of `render` and HMR works, although the whole notebook is reloaded on every hot update.

kwentine:
Fixes #586 

It is a rough starting point inspired from `./anywidget/src/widget.js` helpers, with warnings abridged.


packages/vite/index.js
```diff
@@ -19,15 +19,29 @@ function showErrorOverlay(err) {
...
-import.meta.hot.accept("${src}", (newModule) => {
-	import.meta.hot.data.render = newModule.render;
+async function getRender(newModule) {
+        let newRender == newModule.render;
+        if (newRender) {
+            console.warn('[anywidget] Deprecation Warning. Direct export of a \'render\' will likely be deprecated in the future.')
+            return newRender;
+        }
+        newRender = newModule.default.render;
+	assert(
```
manzt, line 29: Have you tested the plugin locally? I would assume this would raise a `ReferenceError` since `assert` is not a global.

manzt:
> Related to ease-of-testing (among other benefits) I throw in an idea: instead of using a long inline template, would it be worthwhile to extract it in a hmr.js file that assumes the existence of global.__anywidget_hmr_src, and just template a global.__anywidget_hmr_src=<src> header ? The concatenation of hmr.js and header being then served by the plugin.

love the idea! Yes, the HMR stuff has just grown in complexity beyond an inline string. I think we could just find and replace a special string:

```js
// hmr.js
const SRC = "__ANYWIDGET_HMR_SRC__";

import.meta.hot.accept(SRC, (newModule) => {
  // ...
});

export async function render({ model, el } ) {
  if (import.meta.hot.data.render == null) {
    let m = await import(SRC);
    // ...
  }
  // ...
}
```

And simply:

```js
// vite/index.js
import * as fs from "node:fs";
import * as path from "node:path";
import * as url from "node:url";

// ceremony to get the filepath, in node>=v20 import.meta.dirname
let __dirname = path.dirname(url.fileURLToPath(import.meta.url));
let template = fs.readFileSync(path.resolve(__dirname, "hmr.js"), { encoding: "utf-8" });
```

then in `load(id)`

```js
return template.replace("__ANYWIDGET_HMR_SRC__", id.split(":")[1]);
```

packages/vite/index.js
manzt: No need to apologize! I really really appreciate the contribution. Just wanted to make sure.

You can also test out the "local" version without much ceremony by pointing to `index.js`

```js
// vite.config.js
import anywidget from "./path/to/anywidget/packages/vite/index.js";

export default {
  plugins: [anywidget()]
}
```

kwentine:
So I gave it a try. Daunted by the file loading subtleties I relied on the esbuild file loader. And i seemed that Vite's static analysis was puzzled by the `SRC = "__ANYWIDGET_HMR_SRC__` indirection, so I had to inline the assigned string repeatedly instead. Tested locally of course 😇

manzt:
> Daunted by the file loading subtleties I relied on the esbuild file loader. 

Interesting workaround! Unfortunately, we only use esbuild to generate the `.cjs` version of the package (i.e., so folks can `require("@anywidget/vite")` rather than `import` from their `vite.config.js`. We ship `index.js` to NPM directly for use in Node via ESM; that means it's the main entry point for the package and is not intended to be bundled.

The code snippet I shared should work:

```js
import * as fs from "node:fs";
import * as path from "node:path";
import * as url from "node:url";

let __dirname = path.dirname(url.fileURLToPath(import.meta.url));
let template = fs.readFileSync(path.resolve(__dirname, "hmr.js"), { encoding: "utf-8" });
```

> And i seemed that Vite's static analysis was puzzled by the SRC = "__ANYWIDGET_HMR_SRC__"

Ah, I guess that makes sense. Thanks for the workaround!

kwentine:
🤔 I will need to understand the matter better, but in the meantime reverted to your implementation. 

I chose a different route in the first place because I now get this errors:
- A build time warning: `"import.meta" is not available with the "cjs" output format and will be empty`
- An error running `vite serve` from a test project: `Identifier '__dirname' has already been declared`

My `vite.config.js` uses `import anywidget from '@anywidget/vite'` and I have `node_modules/@anywidget/vite -symlink-> path/to/local/repo/anywidget/packages/vite`.

I have the impression that notwithstanding the `import` syntax, `node` loads the built `dist/index.cjs`. Hence the esbuild text loader approach worked for me even though I did not `require('@anywidget/vite')` in `vite.config.js`.

kwentine:
A little idea before the weekend: could we not trigger the builtin anywidget HMR mechanism in response to Vite HMR updates ?
The following snippet could be injected by a `transform` plugin just before `export default { render }`
```js
if (import.meta.hot) {
    
    let SRC = "/index.js"; // "__ANYWIDGET_HMR_SRC__"
    let PORT = "5173";     // "__ANYWIDGET_HMR_PORT__"
    
    // Touch _esm on all active models
    async function refresh() {
	for (let m of import.meta.hot.data.models) {
	    m.set("_esm", `http://localhost:${PORT}${SRC}?t={Date.now()}`);
	}
    }

    let originalRender = render;
    render = async function ({model, el}) {
	import.meta.hot.data.models.add(model);
	return await originalRender({model, el})
    }

    // Register event handlers on first load only
    if (!import.meta.hot.data.models) {
	
	import.meta.hot.data.models = new Set();

	import.meta.hot.on('vite:beforeUpdate', (data) => {
	    for (let u of data.updates) {
		if (u.path === SRC) {
		    return refresh()
		}
	    }
	})

	// Accept HMR update to avoid full page reload
	import.meta.hot.accept((newModule) => {})

    }    
}
```
Be that as it may, I wish you a nice weekend ;)

manzt:
I have not forgotten about this! I've been busy working on #603 and just getting back to this. I'll try to give review and respond to your commends ASAP.

manzt:
> I adopted your suggestion, thanks for the clarification. I think my understanding of the matter is limited, though 🤔

I wonder if we can just go ESM-only. and forget the esbuild stuff. NodeJS supports `require`-ing ESM now, so I think we can move towards ESM-only plugin.

manzt:
> A little idea before the weekend: could we not trigger the builtin anywidget HMR mechanism in response to Vite HMR updates ?

The built-in anywidget HMR will not be enabled when using Vite. Built-in HMR only starts up if a file path is detected for `_esm`, and when you use vite you have a URL:

```py
_esm = "http://localhost:5173/index.js?anywidget"
```

manzt:
Unfortunately, I'm getting a weird console error with the plugin now:

```
TypeError: Failed to fetch dynamically imported module: http://localhost:5173/index.tsx?anywidget
```

when I use a dynamic `import` with Vite.

manzt:
Thanks so much for the contribution! Will cut a release ASAP.

kwentine:
Thanks for your advice and explanation in the process of getting this merged, it was fun and instructive. 👍🏻

PR #609, opened 2024-06-21
chore: Improve ESM deprecation notice info

manzt:
cc: @flekschas


flekschas (approved):
Nice! Thanks for putting this together so quickly. It'll be really nice for debugging.

PR #619, opened 2024-07-06
Add trove classifiers to pyproject.toml

nvictus:
The framework classifiers should make anywidget discoverable within the JLab extension manager.

See [classifiers](https://pypi.org/classifiers/).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,6 +18,16 @@ dependencies =  [
     "typing-extensions>=4.2.0",
     "psygnal>=0.8.1",
 ]
+classifiers = [
+    "Framework :: Jupyter",
+    "Framework :: Jupyter :: JupyterLab",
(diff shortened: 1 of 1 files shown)


manzt:
Do you know the purpose of these classifiers are in general? I'm somewhat doubtful they do much (especially these generic ones), and don't love the idea of them also going out of date over time. Anywidget isn't really a JupyterLab extension also, since it's not meant to be installed in the UI but rather via pip.

nvictus:
They're pretty widely used in Python packages. Their main purpose is for searching and indexing on PyPI. The Framework classifiers are being used specifically by JupyterLab to discover extensions (via the PyPI API). Only minor-version-specific identifiers will go out of date as you deprecate support for older versions and support newer ones (which is why I didn't include any).

PR #621, opened 2024-07-11
Change the property name from count to value

giswqs:
The property name in the example is `value` instead of `count`.


manzt (approved):
Thanks so much for the contribution!

PR #629, opened 2024-07-24
feat: Patch ipywidgets to be more flexible with child widgets

manzt:
## Summary

ipywidgets performs an `isinstance(x, Widget)` check with their container classes (i.e., `Box`, `Link`). This restricts the use of only classes that derive from `ipywidgets.Widget`, which is a very large class with many methods.

This is problematic because it ties the entire widgets ecosystem (and anywidget) to a large, monolithic class. Widget authors inherit from this class, and now their top-level APIs for end users have a set of methods that should be considered private to end users but are useful for widget developers.

This makes it difficult to innovate on Python-side APIs, as developers are forced to inherit from this large class even if they only need basic widget functionality.

In anywidget, it _really_ holds us back from iterating on the Python side of thing because we don't want to break compatability with ipywidgets but we _need_ other APIs (#579, #628).

## Solution

This PR patches ipywidgets to allow widgets composed from non-`Widget` base classes. The patch performs flexible serialization and deserialization by checking for the model_id attribute instead of strict instance checks.

We should probably come up with a _protocol_ for getting the `model_id` attribute, making this even less tied to anywidget/ipywidgets, but for now, this is a good start.

### DIFF
--- a/anywidget/__init__.py
+++ b/anywidget/__init__.py
@@ -2,10 +2,13 @@
 
 from __future__ import annotations
 
+from ._patch_ipywidgets import _patch_ipywidgets
 from ._version import __version__
 from .widget import AnyWidget
 
 __all__ = ["AnyWidget", "__version__"]
+
+_patch_ipywidgets()
 
 
 def _jupyter_labextension_paths() -> list[dict]:
--- a/anywidget/_patch_ipywidgets.py
+++ b/anywidget/_patch_ipywidgets.py
@@ -1,0 +1,104 @@
+"""ipywidgets patch module.
+
+Patches ipywidgets to allow for more flexible serialization and
+deserialization of (any)widgets by allowing objects that are not
+strict instances of `ipywidgets.Widget`.
+
+Only `ipywidgets.Box` and `ipywidgets.widgets.widget_link.Link`
+give problems. This code is mostly vendored from ipywidgets and
+modified to allow for more flexibility.
+"""
+
+from __future__ import annotations
+
+import typing as t
+
+import ipywidgets
+import traitlets
+from ipywidgets.widgets.trait_types import TypedTuple
+from ipywidgets.widgets.widget import Widget, _instances
+
+from ._descriptor import _COMMS
+
+
+def _get_model_id(x: t.Any) -> str | None:
+    """Get the model id of a widget or comm."""
+    if isinstance(x, Widget):
+        return x.model_id
+    maybe_comm = _COMMS.get(id(x), None)
+    return getattr(maybe_comm, "comm_id", None)
+
+
+def _widget_to_json(x: t.Any, obj: t.Any) -> t.Any:
+    """Recursively convert a widget to json."""
+    if isinstance(x, dict):
+        return {k: _widget_to_json(v, obj) for k, v in x.items()}
+    elif isinstance(x, (list, tuple)):
+        return [_widget_to_json(v, obj) for v in x]
+    model_id = _get_model_id(x)
+    return f"IPY_MODEL_{model_id}" if model_id else x
+
+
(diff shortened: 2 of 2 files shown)


manzt:
For example, with this PR, widget authors could extend directly from `traitlets.HasTraits` and work with regular ipywidgets:

```python
import traitlets
import ipywidgets

esm = """
export function render({ model, el }) {
  let count = () => model.get("value");
  btn.classList.add("counter-button");
    model.set("value", count() + 1);
    model.save_changes();
  });
  model.on("change:value", () => {
  });
  el.appendChild(btn);
}
"""

css = """
.counter-button {
  background-color: rgba(148, 163, 184, 0.1);
  border-radius: 10px;
  padding: 0.5em 1em;
  font-size: 2em;
}
"""

class Counter(traitlets.HasTraits):
    _repr_mimebundle_ = MimeBundleDescriptor(_esm=esm, _css=css)

counter = Counter()
ipywidgets.VBox([counter, ipywidgets.IntSlider()])
```

manzt:
@tlambert03, I'm really flummoxed here. I believe it's crucial to maintain compatibility with these top-level APIs from ipywidgets, but the deep connection with ipywidgets.Widget is really holding us back.

I think this is the minimal "patching" that is necessary, but it's definitely a hack. I just worry it's going to take a long time to get some changes like this in ipywidgets and would like to keep iterating here.

tlambert03:
haven't fully looked through your `_patch_ipywidgets` solution yet... but, just a question:  if you could somehow "magically" make some anywidget object pass the `isinstance(x, Widget)` check, would that solve the problem?

manzt:
> if you could somehow "magically" make some anywidget object pass the isinstance(x, Widget) check, would that solve the problem?

Yes... that would solve everything I think (if we also exposed a `model_id` on our class). However, trying to look into how to do that I couldn't figure it out :)

tlambert03:
hmmm, yeah, i was hoping i could be sneaky with `__instancecheck__` https://docs.python.org/3/reference/datamodel.html#customizing-instance-and-subclass-checks ... but running into probably the same difficulties you did.  It's easy to make anything else look like an instance of something you own, but hard to make something you own look like an instance of anything else

tlambert03:
here's another crazy black magic idea:

```python
In [8]: import ipywidgets

In [9]: class MyThing:
   ...:     @property
   ...:     def __class__(self):
   ...:         return ipywidgets.Widget
   ...:

In [10]: isinstance(MyThing(), ipywidgets.Widget)
Out[10]: True
```

so, taking that one step further, if you did a little stack inspection inside of that `__class__` method, and checked "who's asking" ... and lied only to ipywidgets, it could work

tlambert03:
```python
import inspect
from ipywidgets.widgets.widget import _widget_to_json
import ipywidgets


class LieToIpyWidgets:
    def __get__(self, obj, objtype=None):
        if obj is None:
            return self
        stack = inspect.stack()

        for frame in stack:
            if "ipywidgets" in frame.filename:
                return ipywidgets.Widget
        return objtype


class Thing:
    __class__ = LieToIpyWidgets()
    model_id = "1234"


thing = Thing()
print(isinstance(thing, ipywidgets.Widget))
print(_widget_to_json(thing, {}))
```

```
➜ python x.py
False
IPY_MODEL_1234
```

tlambert03:
not saying it's much better than what you're currently doing ... but I do definitely hesitate to monkey patch other peoples libraries, and they would be justified in being annoyed by that

manzt:
> not saying it's much better than what you're currently doing ... but I do definitely hesitate to monkey patch other peoples libraries, and they would be justified in being annoyed by that

Totally. I like your solution a lot better. I think I'm going to go that direction. Thanks so much for the experimentation and thoughts. I'm not sure I would have come to a similar alternative.

PR #630, opened 2024-07-27
Update getting-started.mdx to include marimo

akshayka:
Small docs change to mention that `anywidget` is supported in marimo :)


manzt:
Awesome! Let's get that framework section landed next :)

PR #688, opened 2024-09-20
Fix link in jupyter-widgets-the-good-parts.mdx

dvdkouril:
Randomly found this `./` link in docs. On localhost that redirects correctly, but when deployed, the resulting link url is [https://anywidget.dev/en/jupyter-widgets-the-good-parts/community](https://anywidget.dev/en/jupyter-widgets-the-good-parts/community) which 404s.

Seems to be the only occurence.

### DIFF
--- a/docs/src/pages/en/jupyter-widgets-the-good-parts.mdx
+++ b/docs/src/pages/en/jupyter-widgets-the-good-parts.mdx
@@ -415,7 +415,7 @@ return value is sent to the front end.
 
 A more complex serialization example might be to serialize a `pandas.DataFrame`
 to the [Apache Arrow](https://arrow.apache.org/#) format or a `numpy` array to
-its underlying bytes. Several [community examples](./community) should serve as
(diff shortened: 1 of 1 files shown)


docs/src/pages/en/jupyter-widgets-the-good-parts.mdx
```diff
@@ -415,7 +415,7 @@ return value is sent to the front end.
 
 A more complex serialization example might be to serialize a `pandas.DataFrame`
 to the [Apache Arrow](https://arrow.apache.org/#) format or a `numpy` array to
-its underlying bytes. Several [community examples](./community) should serve as
+its underlying bytes. Several [community examples](/community) should serve as
```
manzt, line 418:
```suggestion
its underlying bytes. Several [community examples](/en/community) should serve as
```

This is vestigial from the template I used, but every path is prefixed with the language (english)

manzt (approved):
Thanks!

PR #780, opened 2025-01-19
chore(docs): Add Numerous under integrations on community page

Lasse-numerous:
Added Numerous apps as an integration in progress to the community page under integrations.

### DIFF
--- a/docs/src/components/IntegrationsGallery.astro
+++ b/docs/src/components/IntegrationsGallery.astro
@@ -19,6 +19,12 @@ let integrations = [
 		logo: "https://raw.githubusercontent.com/executablebooks/MyST-Parser/master/docs/_static/logo-wide.svg",
 		status: "planned",
 	},
+	{
+		name: "Numerous Apps",
+		url: "https://github.com/numerous-com/numerous-apps",
+		logo: "https://raw.githubusercontent.com/numerous-com/numerous-apps/refs/heads/main/logo-with-text.svg",
+		status: "planned",
+	}
 ] satisfies Array<Integration>;
 ---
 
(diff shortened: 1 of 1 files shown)


Lasse-numerous:
Hi @manzt, Hope you'll merge this PR adding Numerous apps project as an integrator of AnyWidget.

Basically the project is a Python  app framework build around using AnyWidgets as its reactive UI components - no proprietary framework standard for widgets. I did this because I wanted make a framework that gives the user full control over the UX in a simple way.

I like the separation of concerns between the app framework and the UX itself which AnyWidget makes possible.

I applied the "in progress" tag because my framework is pretty new and evolving fast still.

manzt:
Thanks for the contribution and integrating anywidget with Numerous.

Do you have a quickstart guide or example repo available that doesn’t require signing up for the service? I’d love to see an example and get a feel for how it works in practice.

Also, just a quick note: we try to keep the canonical naming of anywidget consistent (all lowercase). Feel free to update references to align with this!

docs/src/components/IntegrationsGallery.astro
```diff
@@ -19,6 +19,12 @@ let integrations = [
 		logo: "https://raw.githubusercontent.com/executablebooks/MyST-Parser/master/docs/_static/logo-wide.svg",
 		status: "planned",
 	},
+	{
+		name: "Numerous Apps",
+		url: "https://github.com/numerous-com/numerous-apps",
+		logo: "https://raw.githubusercontent.com/numerous-com/numerous-apps/refs/heads/main/logo-with-text.svg",
+		status: "planned",
+	}
```
manzt, line 27:
```suggestion
	},
```

manzt: To fix the linting error

Lasse-numerous: Thank you for fixing

Issue #803, opened 2025-02-21
Early comm messages to widget are lost


iisakkirotko:
### Describe the bug

When, for example, widget traits are changed quickly after the creation of the widget, these updates would not be received by the widget. This happens because the widget model is loaded asynchronously by `solid.createResource` at https://github.com/manzt/anywidget/blob/ce7f298dec151701ee20c37db09ebadd0bd4154f/packages/anywidget/src/widget.js#L336-L340

Any comm messages that are sent before `load_widget` resolves are lost.

### Reproduction

Take a look at [this PyCafe example](https://py.cafe/iisakkirotko/solara-anywidget-integration). If you open the developer console, you can see that none of the change-triggered console messages are sent.

I also think that https://github.com/plotly/plotly.py/issues/4933 (which was fixed differently in https://github.com/plotly/plotly.py/pull/4956) was ultimately due to this behaviour.

### Logs

```shell

```

### System Info

<details>
<summary>System info</summary>

```shell
System:
    OS: macOS 15.3
    CPU: (10) arm64 Apple M2 Pro
    Memory: 909.98 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
Browsers:
    Chrome: 133.0.6943.127
    Safari: 18.3
    Safari Technology Preview: 18.2
```

```shell
-e file:///Users/iisakki/Code/iisakkirotko/anywidget
jupyter-client==8.6.3
jupyter-core==5.7.2
jupyter-events==0.12.0
jupyter-lsp==2.2.5
jupyter-server==2.15.0
jupyter-server-terminals==0.5.3
jupyterlab==4.3.5
jupyterlab-pygments==0.3.0
jupyterlab-server==2.27.3
jupyterlab-widgets==3.0.13
notebook==7.3.2
notebook-shim==0.2.4
```

</details>

### Severity

annoyance

PR #804, opened 2025-02-21
fix: comm messages were lost before model was loaded

iisakkirotko:
As I explained in the related issue https://github.com/manzt/anywidget/issues/803, any comm messages sent to the widget before it's model was loaded were lost. This should ensure that those messages are also processed.

Fixes: https://github.com/manzt/anywidget/issues/803.

### DIFF
--- a/packages/anywidget/src/widget.js
+++ b/packages/anywidget/src/widget.js
@@ -309,53 +309,58 @@ class Runtime {
 	/** @type {import('solid-js').Resource<Result<AnyWidget & { url: string }>>} */
 	// @ts-expect-error - Set synchronously in constructor.
 	#widget_result;
+	/** @type {Promise<void>} */
+	model_created;
 
 	/** @param {base.DOMWidgetModel} model */
 	constructor(model) {
-		this.#disposer = solid.createRoot((dispose) => {
-			let [css, set_css] = solid.createSignal(model.get("_css"));
-			model.on("change:_css", () => {
-				let id = model.get("_anywidget_id");
-				console.debug(`[anywidget] css hot updated: ${id}`);
-				set_css(model.get("_css"));
-			});
-			solid.createEffect(() => {
-				let id = model.get("_anywidget_id");
-				load_css(css(), id);
-			});
+		this.model_created = new Promise((resolve) => {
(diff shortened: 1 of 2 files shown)


packages/anywidget/src/widget.js
```diff
@@ -309,53 +309,58 @@ class Runtime {
...
-				load_css(css(), id);
-			});
-
-			/** @type {import("solid-js").Signal<string>} */
-			let [esm, setEsm] = solid.createSignal(model.get("_esm"));
-			model.on("change:_esm", async () => {
-				let id = model.get("_anywidget_id");
-				console.debug(`[anywidget] esm hot updated: ${id}`);
-				setEsm(model.get("_esm"));
+		this.model_created = new Promise((resolve) => {
```
manzt, line 317: May I suggest using [`Promise.withResolvers`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/withResolvers) to help with readability here?

```ts
let { promise, resolve } = Promise.withResolvers();
this.ready = promise;
this.#disposer = solid.createRoot((dispose) => {
  /* ... */ 
  resolve();
  /* ... */ 
});
```

packages/anywidget/src/widget.js
```diff
@@ -309,53 +309,58 @@ class Runtime {
 	/** @type {import('solid-js').Resource<Result<AnyWidget & { url: string }>>} */
 	// @ts-expect-error - Set synchronously in constructor.
 	#widget_result;
+	/** @type {Promise<void>} */
+	model_created;
```
manzt, line 313: Slight preference to rename this to something like `Runtime.ready`, in case there is other setup in the future that is required. e.g.,

```ts
const runtime = RUNTIMES.get(this);
await runtime?.ready;
```

manzt (approved):
Wow, thanks for the quick fix. Thank you! Just a couple of minor nits.

manzt (approved):
Wow, thanks for the quick fix. Thank you! Just a couple of minor nits.

iisakkirotko:
Should look better now!

manzt:
Slick, thanks so much for the contribution. I'll cut a release now.

PR #854, opened 2025-05-09
feat: Vue AFM bridge

aryan02420:
Hey, awesome project! 🎉

I wanted to create some custom widgets using Vue, and since you already have packages for React and Svelte, I decided to contribute the Vue binding code.

The API of the Vue binding is very similar to the @anywidget/react package. The major difference is that instead of returning a getter-setter pair from `useModelState`, we return a `ShallowRef` which can be mutated directly (eg. `value++`). This is more idiomatic to Vue's reactivity system.

## Current Status

I haven't figured out how to get HMR to work, so for now, you’ll need to restart the kernel for updated JS/CSS files to be picked up. If this can be addressed in a patch release, great! Otherwise, I’ll revisit it in a few days to troubleshoot further.

## Demo

You can check out a demo of this package on the branch [aryan02420:demo/vue-bridge](https://github.com/aryan02420/anywidget/tree/demo/vue-bridge/packages/vue-demo).

demo of the custom widget rendered using vue

### DIFF
--- a/packages/vue/CHANGELOG.md
+++ b/packages/vue/CHANGELOG.md
@@ -1,0 +1,7 @@
+# @anywidget/vue
+
+## 0.0.1
+
+### Patch Changes
+
+- fix: specify "types" field in package.json ([#198](https://github.com/manzt/anywidget/pull/198))
--- a/packages/vue/CHANGELOG.md
+++ b/packages/vue/CHANGELOG.md
@@ -1,7 +1,1 @@
 # @anywidget/vue
-
-## 0.0.1
-
-### Patch Changes
-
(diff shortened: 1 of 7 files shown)


manzt:
Thank you very much for the contribution. I have been on vacation and traveling and will be able to take a look this week. I really appreciate your patience in my review.

manzt (approved):
Wow, thanks so much for the contribution! I'm not super familiar with the idiomatic Vue patterns so I really appreciate the thought that went into this. Tried it out and things worked really well.

I agree, we can try to sort out HMR in a separate PR!

PR #857, opened 2025-05-21
chore: codespell

schwehr:
Would be good to setup a codespell pre-commit, but just do a run by hand for now.

```
codespell --version
2.4.1
```


manzt:
> Would be good to setup a codespell pre-commit, but just do a run by hand for now.

Thanks for the contribution! We don't use pre-commit currently, but certainly something to consider if we do.

PR #881, opened 2025-08-18
Normalize possible AFM inputs to an AFM object when using Vite HMR.

GuillaumeQuenneville:
This PR changes @anywidget/vite to follow the AFM spec in hmr.js, see [docs](https://anywidget.dev/en/afm/). Supported inputs are a default exported AFM object or callable that returns an AFM object (useful if closures are needed between initialize and render. Added two legacy exports and give a console.warn() about deprecation.

The previous behaviour handled exporting render functions but not AFM objects. @anywidget/react was not following the spec, in that the examples would return a render callable instead of a callable to an AFM object. Updating the react behaviour is TODO, I didn't have a look at the other plugins. See below for react case.
https://github.com/manzt/anywidget/blob/25454977424a29a5fcadf82c5a9536ff9eb67fbb/packages/react/index.js#L88-L108

This PR
- normalizes all possible export strategies when using HMR to an AFM object on best effort basis.
- This fixes Svelte Vite HMR when using defineWidget from @anywidget/svelte which returns a callable to an AFM object.
- print a console warning when using deprecated export strategies.

I tested all cases manually with following with and without an initialize.
/**
 * Normalizes possible AFM inputs to an AFM object: { initialize?, render }. Accepts callables to provide closures.
 * 1) (deprecated): exported `render` and `initialize?` -> AFM object
 * 2) AFM object
 * 3) (deprecated): Callable -> callable render -> AFM object
 * 4) Callable -> AFM object
 * @param newModule the module
 * @returns normalize AFM object
 */
I don't know what is recommended to write tests for HMR plugins. I'm a python/C++ dev so it's pretty deep in the DX stack for me.


packages/vite/hmr.js
```diff
@@ -64,7 +103,8 @@ async function refresh() {
 		}
 		context.model.off();
 		emptyElement(context.el);
-		let cleanup = await render({ model: context.model, el: context.el });
+		afm.initialize?.({ model: context.model });
+		let cleanup = await afm.render({ model: context.model, el: context.el });
```
manzt, line 107: Both `initialize` and `render` can optionally return a cleanup function. I think we'd want to run those in sequence:

```ts
		let cleanupIntialize = await afm.initialize?.({ model: context.model });
		let cleanupRender = await afm.render({
			model: context.model,
			el: context.el,
		});
		context.cleanup = () => {
			cleanupIntialize?.();
			cleanupRender?.();
		};
```

manzt (commented):
Thanks for the contribution! As you’ve noticed, the Vite plugin hasn’t seen much attention since we implemented our own HMR in anywidget. I had been considering suggesting esbuild/rolldown with our HMR, but the `@tailwindcss/vite` plugin you shared on Discord doesn’t seem to work out-of-the-box with rolldown. There is a third-party Rollup plugin that integrates with Tailwind and our native HMR, but since many folks will likely prefer to stick with Vite + Tailwind, this is a really valuable addition.

packages/vite/hmr.js
GuillaumeQuenneville: Thanks for the feedback

manzt:
Thanks for the contribution!

PR #1002, opened 2026-06-22
fix: require svelte widget props in types

### LINKED ISSUE #1000: Incorrect typing in `@anywidget/svelte`
### Describe the bug

The `defineWidget` function expects as input a Svelte component of type `svelte.Component<{ model?: AnyModel<T>, bindings?: T }>`. This means the `$props` of the component must be typed to accept `undefined` for `model` and `bindings`. This doesn't make sense, considering both are always passed to the component in reality. The practical downside is that it requires either explicit handling of `undefined` values, or explicit type cast.

If I understand correctly, the intention was to make the parameters optional. If so, the `?` have no effect on that with the idiomatic way of typing the props (`let { [extracted parameters] }: [type] = $props();` -- this only specifies type restrictions for the affected subset of props, allowing arbitrary extras). Changing the expected type to `svelte.Component<{ model: AnyModel<T>, bindings: T }>` will still allow omitting `model` or `bindings` from the prop declarations, will remain compatible with any existing components allowing `undefined`, and will simplify the typing for newly written widgets.

### Reproduction

```svelte
// Widget.svelte
<script lang="ts">
    interface Props {
        bindings: {
            foo: number;
            bar: string;
        };
    };
    
    let { bindings }: Props = $props();
</script>
```

```typescript
// index.ts
import { defineWidget } from "@anywidget/svelte";
import Widget from "./Widget.svelte";

export default defineWidget(Widget);
/*
Results in an error:
Argument of type 'Component<Props, {}, "">' is not assignable to parameter of type 'Component<{ model?: AnyModel<Bindings> | undefined; bindings?: Bindings | undefined; }, {}, string>'.
  Types of parameters 'props' and 'props' are incompatible.
    Type '{ model?: AnyModel<Bindings> | undefined; bindings?: Bindings | undefined; }' is not assignable to type 'Props'.
      Types of property 'bindings' are incompatible.
        Type 'Bindings | undefined' is not assignable to type 'Bindings'.
          Type 'undefined' is not assignable to type 'Bindings'.ts(2345)
*/
```

Currently, fixing the type error requires changing `bindings: {` to `bindings?: {` in `interface Props`, which does not match the conceptual contract of a widget.

### Severity

annoyance

sanmaxdev:
## Summary
- Update the `@anywidget/svelte` `defineWidget` component type so `model` and `bindings` are required props.
- Add a runtime guard before mounting to keep the required `bindings` type sound.

Closes #1000

## Testing
- [x] `pnpm --filter @anywidget/svelte exec tsc --noEmit false --emitDeclarationOnly`
- [x] `pnpm exec tsc --build packages/svelte/tsconfig.json --pretty false`
- [x] `pnpm exec vp check packages/svelte/src/index.js`

### DIFF
--- a/.changeset/serious-birds-wink.md
+++ b/.changeset/serious-birds-wink.md
@@ -1,0 +1,5 @@
+---
+"@anywidget/svelte": patch
+---
+
+Tighten Svelte widget component props so `model` and `bindings` are required.
--- a/packages/svelte/src/index.js
+++ b/packages/svelte/src/index.js
@@ -50,7 +50,7 @@ function createBindings(model) {
  * ```
  *
  * @template {Record<string, any>} T
- * @param {svelte.Component<{ model?: AnyModel<T>, bindings?: T }>} Widget
+ * @param {svelte.Component<{ model: AnyModel<T>, bindings: T }>} Widget
  * @returns {AnyWidget<T>}
  */
 export function defineWidget(Widget) {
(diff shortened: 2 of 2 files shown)


manzt (approved):
So sorry for the delay on this, it looks great.

